PR #1, opened 2016-07-09
Updated spec file with new git changes, proper package dependencies, etc

dranch:


tvrusso:
Hmmm.  Does the URL in the spec file for the archive actually work?  We have not started using github releases, and the latest released source tarball is still on sourceforge.

Our release tarballs have traditionally been bootstrapped, and as far as I can tell github releases are just tarballs of tagged repo states.  I am not yet sure it's possible to bundle bootstrapped versions into a github release.

Did this spec really work on CentOS as is?

tvrusso:
While I can't review the actual validity of this RPM spec for CentOS, it does appear that the updated URL does correspond to what github provides as "release" tarballs, and since the spec DOES run bootstrap.sh, it doesn't matter that github tarballs don't have bootstrap run, as our sourceforge release tarballs did.

In addition, the spec adds prerequisite packages (which the older one lacked), and uses more sane configure options.

PR #6, opened 2016-11-05
Enable IPV6 support for TCP server.

dustin:
Hi,

I was connecting to my xastir server from APRSDroid on a tablet and then tried it on my phone and was getting connection refused.  After a bit of investigation, I figured out that the problem was that android N is trying to connect using IPV6 and the xastir server is bound specifically on IPV4.

With this change, both work fine.

### DIFF
--- a/src/x_spider.c
+++ b/src/x_spider.c
@@ -920,9 +920,28 @@ void set_proc_title(char *fmt,...) {
     Argv[1] = ((void *)0) ;
 }
 
+#define ADDR_STR_LEN 39
 
+char *addr_str(const struct sockaddr *sa, char *s) {
+    switch(sa->sa_family) {
+    case AF_INET:
+        inet_ntop(AF_INET, &(((struct sockaddr_in *)sa)->sin_addr),
+                  s, ADDR_STR_LEN);
+        break;
 
+    case AF_INET6:
+        inet_ntop(AF_INET6, &(((struct sockaddr_in6 *)sa)->sin6_addr),
+                  s, ADDR_STR_LEN);
+        break;
 
+    default:
+        xastir_snprintf(s, ADDR_STR_LEN, "<unknown family: %d>",
+                        sa->sa_family);
+        return NULL;
+    }
+
+    return s;
+}
 
 // This TCP server provides a listening socket.  When a client
 // connects, the server forks off a separate process to handle it
@@ -958,17 +977,17 @@ void TCP_Server(int argc, char *argv[], char *envp[]) {
 
     int sockfd, newsockfd, childpid;
     socklen_t clilen;
-    struct sockaddr_in cli_addr, serv_addr;
+    struct sockaddr_in6 cli_addr, serv_addr;
     pipe_object *p;
     int sendbuff;
     int pipe_to_parent; /* symbolic names to reduce confusion */
     int pipe_from_parent;
     char timestring[101];
+    char addrstring[ADDR_STR_LEN+1];
 
-    
     // Open a TCP listening socket
     //
-    if ( (sockfd = socket(AF_INET, SOCK_STREAM, 0)) < 0) {
+    if ( (sockfd = socket(AF_INET6, SOCK_STREAM, 0)) < 0) {
         fprintf(stderr,"x_spider: Can't open socket for listening\n");
         fprintf(stderr,"Could some processes still be running from a previous run of Xastir?\n");
         exit(1);
@@ -997,9 +1016,9 @@ void TCP_Server(int argc, char *argv[], char *envp[]) {
     // Bind our local address so that the client can send to us.
     //
     memset((char *)&serv_addr, 0, sizeof(serv_addr));
-    serv_addr.sin_family = AF_INET;
-    serv_addr.sin_addr.s_addr = htonl(INADDR_ANY);
-    serv_addr.sin_port = htons(SERV_TCP_PORT);
+    serv_addr.sin6_family = AF_INET6;
+    // serv_addr.sin6_addr.s6_addr = htonl(INADDR_ANY);
+    serv_addr.sin6_port = htons(SERV_TCP_PORT);
 
     if (bind(sockfd,
             (struct sockaddr *)&serv_addr,
@@ -1045,7 +1064,8 @@ void TCP_Server(int argc, char *argv[], char *envp[]) {
                 // no incoming socket connection.  Check the pipe
                 // queues for incoming data.
                 //
-                if (pipe_check(inet_ntoa(cli_addr.sin_addr)) == -1) {
+                if (pipe_check(addr_str((struct sockaddr*)&cli_addr,
+                                        addrstring)) == -1) {
 
                     // We received a shutdown command from the
                     // master socket connection.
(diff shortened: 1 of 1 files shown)


tvrusso:
I tried out this change, and with it I can no longer connect to a running Xastir's server port (on the machine with the change) from another xastir instance without it.  That is, I can no longer use server ports between machines using IPv4.

Can you test out your changes and make any modifications necessary to maintain proper operation using IPv4 while still providing IPv6 service?

dustin:
Hmm...  I definitely connected via IPv4 before I sent that.  I'm not
entirely sure how your test ran.

pi@raspberryeye:~ $ nc 192.168.1.111 2023
# Welcome to Xastir's server port, callsign: KG6HWF
^C
pi@raspberryeye:~ $ nc 2602:306:bcef:ac10:241c:d1b7:8de4:c5d 2023
# Welcome to Xastir's server port, callsign: KG6HWF
^C

I'll investigate more.  There might be something in how xastir connects
that makes it not work with IPV6 for some reason.

tvrusso:
OK, so here's what I did to test this out:

I built Xastir from your dustin-ipv6 branch, and ran it on one machine.

I ran Xastir built from master on a separate machine, and told it to connect to the server port of the other machine. Got a connection failure.

When I rebuilt Xastir on the first machine from master, the second machine could connect to it Just Fine.

So there is something not quite right yet.

dustin:
I'm having trouble reproducing this.  :(  I'm able to connect to mine from both master and the package on my raspberry pi.

xastir

This is happily receiving updates.  Can you tell me more about the nature of the failure you're getting?  tcpdump maybe?

tvrusso:
I'll investigate more tonight and post the results.

jpronans:
Hi Tom, dustin, 

It seems to work here
Sun Nov 06 18:52:52 GMT 2016 X_spider client connected from address 2001:770:XXXXX
Sun Nov 06 18:52:53 GMT 2016 X_spider session terminated, unauthenticated user, address 2001:770:XXXXX
Sun Nov 06 18:52:55 GMT 2016 X_spider client connected from address ::ffff:192.168.1.253
Sun Nov 06 18:52:57 GMT 2016 X_spider session terminated, unauthenticated user, address ::ffff:192.168.1.253

Though you probably should do UDP as well.
John
EI7IG

tvrusso:
Ok, so more data.

I can build Xastir with the ipv6 branch code on my Linux laptop and connect to it from my FreeBSD desktop machine (my primary) running master branch code.  It's the other direction that doesn't work, get a connection refused.

I can connect to the FreeBSD/ipv6-branch server port from the FreeBSD machine by telnet to localhost 2023, but cannot do the same from the Linux machine.

So there's something non-portable about the solution.

tvrusso:
K.  The only reason that the BSD machine could talk to itself via "telnet localhost 2023" is that localhost has an ipv6 definition, and it's connecting via that.

If I do "telnet 10.0.0.1 2023" (the ipv4 address of the BSD machine) it doesn't connect.  Connection refused, as if nothing is listening on that port.  

This network doesn't support ipv6, none of the machines in the house are using ipv6 addressing, and the local DNS server is not setup to support it.

tvrusso:
Last comment.

I can do the same experiment on the linux laptop.  Build the ipv6 branch code, then try:

  telnet ::1 2023
works fine, as does
  telnet 127.0.0.1 2023

There is something fishy about the way BSD is listening using the code as supplied.  It's listening only on the ipv6 address with the modified code, and only for ipv4 connections with master.


A quick look at "man inet6" on the BSD machine shows this:

   Interaction between IPv4/v6 sockets
     By default, FreeBSD does not route IPv4 traffic to AF_INET6 sockets.  The
     default behavior intentionally violates RFC2553 for security reasons.
     Listen to two sockets if you want to accept both IPv4 and IPv6 traffic.
     IPv4 traffic may be routed with certain per-socket/per-node configura-
     tion, however, it is not recommended to do so.  Consult ip6(4) for
     details.

     The behavior of AF_INET6 TCP/UDP socket is documented in RFC2553.  Basi-
     cally, it says this:
     o   A specific bind on an AF_INET6 socket (bind(2) with an address speci-
         fied) should accept IPv6 traffic to that address only.
     o   If you perform a wildcard bind on an AF_INET6 socket (bind(2) to IPv6
         address ::), and there is no wildcard bind AF_INET socket on that
         TCP/UDP port, IPv6 traffic as well as IPv4 traffic should be routed
         to that AF_INET6 socket.  IPv4 traffic should be seen as if it came
         from an IPv6 address like ::ffff:10.1.1.1.  This is called an IPv4
         mapped address.
     o   If there are both a wildcard bind AF_INET socket and a wildcard bind
         AF_INET6 socket on one TCP/UDP port, they should behave separately.
         IPv4 traffic should be routed to the AF_INET socket and IPv6 should
         be routed to the AF_INET6 socket.

     However, RFC2553 does not define the ordering constraint between calls to
     bind(2), nor how IPv4 TCP/UDP port numbers and IPv6 TCP/UDP port numbers
     relate to each other (should they be integrated or separated).  Imple-
     mented behavior is very different from kernel to kernel.  Therefore, it
     is unwise to rely too much upon the behavior of AF_INET6 wildcard bind
     sockets.  It is recommended to listen to two sockets, one for AF_INET and
     another for AF_INET6, when you would like to accept both IPv4 and IPv6
     traffic.

     It should also be noted that malicious parties can take advantage of the
     complexity presented above, and are able to bypass access control, if the
     target node routes IPv4 traffic to AF_INET6 socket.  Users are advised to
     take care handling connections from IPv4 mapped address to AF_INET6 sock-
     ets.

Googling:

http://stackoverflow.com/questions/1618240/how-to-support-both-ipv4-and-ipv6-connections

It appears that this issue is not limited to FreeBSD.

jpronans:
Came across this http://www.ipv6forum.com/ipv6_enabled/DNS.php

tvrusso:
Yeah.  So, the short story is that I can't merge this pull request as is.  It would break Xastir in some cases.

As provided, it depends on system-specific default handling of AF_INET6 sockets.  Linux (or at least, some distros of Linux) default to letting them be dual-stack by default, and other OSen (BSD and Windows, at least) don't.  Some OSen would allow setsockopts calls to turn that behavior on, but Windows doesn't.  So the portable solution is harder.

If you want to spend the time making this portable, I'm perfectly willing to test it out on the environments that matter to me.  In the meantime, though, I'm unable to assist with implementing it.

dustin:
Lack of standards compliance is unfortunate.  It's a lot more work to
retrofit for multiple serving sockets.  xastir as shipped is broken for me
today, but I don't want to maintain a patch forever, so I'll try to see
what else works.

godfreja:
I decided to try updating the x_spider TCP code to support multiple sockets of different address families. I still want to do more work and testing before starting a pull request, but if anyone wants to look at the work in progress you can find it at https://github.com/godfreja/Xastir/tree/ipv6server. It seems to work on my Mac.

tvrusso:
Jason:  If you want to get a IPv4/IPv6 fix going that does not depend on IPv4 getting automatically handled by the IPv6 server, I will happily test out any pull request you make.  The issues with this pull request are easy to reproduce on my home LAN, and I can test your your alternate fix there just fine.

If you get one that doesn't have the problems that this one has, I'll merge it and close this one.

Thanks for being willing to put in the time to look into it.

tvrusso:
Pull request #11 fixes this issue in a way that does not rely on the dual-stacking that Linux enables by default (and that some systems consider a security problem, so don't enable even though it's a standard specified by RFC2553).  I merged that just now, rendering this pull request moot.

I am therefore closing this pull request without merging it.

PR #8, opened 2017-08-24
Fix broken uncompressed weather alerts

K2DLS:


tvrusso:
Have you verified that the code with this fix applied still decodes the weather alerts that Pete Loveall is putting onto APRS-IS and that Curt is putting out onto Firenet, as well as having checked it with the test data.

K2DLS:
Curt's work perfectly.  Pete's work with a caveat.  Pete is not following his own standard and sends each alert twice.  One is flagged as compressed even though it uncompressed.  The properly formatted compressed alert displays fine, so the broken one is just a cosmetic issue.

K2DLS:
screenshot from 2017-08-24 17-55-57

Here's a screenshot with a some of Pete's alerts.  You'll see that the first four are flagged with the ? (initial state) but that the compressed version are displayed underneath with the N flag (active outside viewport).  For me, the benefits of uncompressed alerts now working properly outweigh the cosmetics.  No data is lost and functionality is improved.

tvrusso:
I have pulled a copy of your branch down to my local machine and am going to run it for a few days in order to verify that all is still working as expected with a firenet and APRS-IS feed.  

After a few days of reviewing, if all is well I'll merge this pull request.

tvrusso:
There appears to be an issue with your local git configuration --- your user name and email address don't appear to match what GitHub recognizes, and so they're showing up as coming from root at your machine.  This is not so much a problem, it's more of an oddity that will prevent you from getting "credit" in the git history --- your commits will show up as having been done by "root" not K2DLS.

There are ways to fix this, but they take a little caution, and it's best done on your end, in your fork.

Any chance you could review https://help.github.com/articles/changing-author-info using the same user name and email address that GitHub has for you, and apply them to your alert branch, then force push as instructed in that article?  If you do that then your commits will show up in the official project repo with you as author.  

If you force push your alert branch, it should automatically update this pull request, and then I can fix things up on my end before doing the merge.

If you aren't comfortable doing what is suggested in that GitHub article, please email me via my email address on the Xastir mailing list and let me know what email address to use (the one that GitHub has recorded for you), and I can do it on this end in my clone of your branch before I merge.

K2DLS:
No problem Tom, I will address it.  I also have been going over the code and am just testing now an even better fix that addresses the cosmetic and Pete's dupes.  I'm adding a test for any alert that has the "-" but is missing an underscore in title[0] -- which would mean it is actually compressed.  It then gets put through the existing compressed code.  It is being tested now and looking good.

If all checks out I'll commit it and fix the author situation.  I know why that happened -- but thanks for the heads up.  Blame it on sudo...

tvrusso:
This pull request was incorrectly closed, because I incorrectly referenced it in the comment to #11

PR #11, opened 2017-11-07
Add support for IPv6 to the x_spider TCP server

godfreja:
Dustin Sallings had submitted an earlier pull request adding IPv6 support to the TCP server. The request relied on the OS mapping IPv4 requests to an IPv6 socket, which some OS's such as FreeBSD do not support. 

This request takes a different approach of creating more than one socket and using poll() to determine when a connection is pending. This should avoid the issue seen with the previous pull request. I did make use of Dustin's addr_str() function. 

I have tested this on my Mac and it accepts connections via both IPv4 and IPv6. Two sockets are listed as listening. I do not have easy access to test on other platforms at the moment.

I may add UDP support later, but that will be a different pull request. (I like smaller code changes.)

### DIFF
--- a/src/x_spider.c
+++ b/src/x_spider.c
@@ -126,6 +126,7 @@
 #include <fcntl.h>
 #include <string.h>
 
+#include <poll.h>
 #include <netinet/in.h>     // Moved ahead of inet.h as reports of some *BSD's not
                             // including this as they should.
 #include <arpa/inet.h>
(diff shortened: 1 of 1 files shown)


tvrusso:
Thank you.  I have tested this out and it works very well on BSD and Linux.  Since you have tested on Mac, I'm going to conclude that we have a winner.

tvrusso:
This commit referenced the wrong pull request.  It should have referenced #6

PR #13, opened 2017-11-17
README.CYGWIN packages dependencies required update

K2DLS:
I attempted a compile and install of the current git source under Cygwin, trying to reference the list of dependencies in README.CYGWIN.  The dependencies did not align with the current packages offered by Cygwin.  Packages were renamed, replaced or eliminated.  GDAL is now directly supported by Cygwin.  The documentation has been updated accordingly based upon several actual compiles/installs of Xastir under Cygwin on Windows 10.  This was tested under x64.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -17,6 +17,7 @@
    DK7IN    Rolf Bleher
    HI8GN    Jose R Marte A
    IK0YUP   Alessandro Frigeri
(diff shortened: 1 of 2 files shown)


tvrusso:
Thank you for reviewing and updating the CYGWIN documentation.

PR #19, opened 2018-01-21
Add IPv6 APRS-IS Support

godfreja:
Besides the changes to socket creation and connection, a forked
version of getaddrinfo was created. This was done to mimic the
current method of running gethostbyname in a seperate process.


tvrusso:
I have merged this pull request.  When I did so, I had to do an interactive rebase to change the formatting of the commit message (and nothing else) in commit 01899183.  Because what I pushed to the Xastir repo does not exactly match what Jason had in the pull request (the difference in the commit message), github doesn't automatically close the pull request.  I'm closing it manually now.

PR #30, opened 2018-02-19
Allow user-defined packets to be sent to APRS-IS

DL7AD:
Hi,

I'm running some tests and use the APRS packet type "User-defined" therefore. Xastir however doesn't relay those packets to the Internet servers.

73 Sven DL7AD

### DIFF
--- a/src/db.c
+++ b/src/db.c
@@ -17546,7 +17546,6 @@ void decode_info_field(char *call,
         //
         if (!done && !ignore) {         // Other Packets        [APRS Reference, chapter 19]
             done = data_add(OTHER_DATA,call,path,message-1,from,port,origin,third_party, station_is_mine, 0);
-            ok_igate_net = 0;           // don't put data on internet       ????
             if (debug_level & 1)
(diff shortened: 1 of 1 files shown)


tvrusso:
I'm inclined to accept this pull request, because the exclusion of this data from the internet feeds seems to be a somewhat arbitrary decision (by someone no longer associated with Xastir development) dating back over 16 years, and the change is so minimal to fix it.  But need a little more information before I do it.

My primary concern is that we be doing something consistent with other programs and with the general expectations on APRS-IS.  I assume that you're saying that Xastir is the only APRS program you've encountered that does NOT forward this data, and that you're able to inject it routinely into APRS-IS via other programs.  Is that correct?  If there are real-world examples of programs that are already injecting this user-defined data packets into APRS-IS, I really can't see any reason NOT to merge this request.

we7u:
I think I'd go with: if it is in the spec to allow it, send it on through.
It is in the spec, so if individual programs can handle it they are not
compliant.

Curt, WE7U
http://wetnet.net/~we7u
http://www.sarguydigital.com

tvrusso:
Fair enough.  I'm just gonna merge it, then.

PR #34, opened 2018-05-09
update OSM README and add map tile info

halmueller:
Remove references to no-longer-available tile sources. Add description of web Mercator projection and tiling scheme, with Wikipedia links. Correct minor punctuation errors. Note: Cloudmade has been out of the map tile business for at least 5 years.


tvrusso:
I have reviewed this change and it looks reasonable to me, so I'm just going to merge it.

Issue #23, opened 2018-02-16
All map processing in Xastir that uses graphicsmagick uses deprecated "GetIndexes" function.


tvrusso:
GraphicsMagick deprecated the GetIndexes function in January of 2018:
http://www.graphicsmagick.org/Changelog.html

As it says in the changelog above:

magick/pixel_cache.h: Mark GetPixels(), GetIndexes(), and GetOnePixel() as deprecated. Compilers may produce a warning if these functions are used.

I have already encountered some compilers that produce such a warning.

That code should be rewritten to do something other than use GetIndexes.  This may take some investigating to find the appropriate substitute.

tvrusso:
A really quick peek at the "pixel_cache.h" file from GraphicsMagick suggests that GetIndexes is deprecated in favor of AccessMutableIndexes, which has the same argument and return type.  This *might* be a drop-in replacement.  The only question is how long ago did they introduce the latter function, and do we have to play stupid ifdef games to deal with ancient installs of GraphicsMagick, like we used to do for \<hock>\<spit>ImageMagick\</spit>\</hock>?

tvrusso:
A quick googling shows that GraphicsMagick has had the AccessMutableIndexes function for over 6 years.  That makes it extremely likely that we don't have to worry about backward compatibility to handle people using versions too old to support it.

ChocolateFrogsNuts:
It's also deprecated on ImageMagick > 669 (possibly earlier).

I'm currently working on bringing a feature I did a few years ago and never got around to sharing into the current version, and part of that is a set of changes for ImageMagick 6.9.10.
Currently at https://github.com/ChocolateFrogsNuts/Xastir/commits/DLM
I'll throw a pull request in the near future.

tvrusso:
This issue was resolved by pull request #40.

PR #40, opened 2018-07-09
Download Manager

ChocolateFrogsNuts:
Some bug fixes and a Download Manager to allow parallel background downloading of OSM tiles.
Really speeds up map display on high-latency connections (eg satellite internet)

### DIFF
--- a/src/dlm.c
+++ b/src/dlm.c
@@ -1,8 +1,8 @@
 /* -*- c-basic-offset: 4; indent-tabs-mode: nil -*-
  *
- * $Id: dlm.c,v 1.10 2011/12/24 02:24:27 jedunmire Exp $
+ * $Id: dlm.c,v 1.10 2018/07/14 21:32:43 MikeNix Exp $
  *
- * Copyright (C) 2010 The Xastir Group
+ * Copyright (C) 2018 The Xastir Group
  *
  * This file was contributed by Mike Nix.
  *
@@ -22,10 +22,6 @@
  *
  * Look at the README for more information on the program.
  *
- * This file derived from map_OSM.c, tile_mgmt.c and fetch_remote.c which
- * have the following copyrights:
- *    Copyright (C) 1999,2000  Frank Giannandrea
- *    Copyright (C) 2000-2010  The Xastir Group
  */
 
 
--- a/src/dlm.c
+++ b/src/dlm.c
@@ -1,0 +1,930 @@
+/* -*- c-basic-offset: 4; indent-tabs-mode: nil -*-
+ *
+ * $Id: dlm.c,v 1.10 2011/12/24 02:24:27 jedunmire Exp $
+ *
+ * Copyright (C) 2010 The Xastir Group
+ *
+ * This file was contributed by Mike Nix.
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; either version 2
+ * of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
+ *
+ * Look at the README for more information on the program.
+ *
+ * This file derived from map_OSM.c, tile_mgmt.c and fetch_remote.c which
+ * have the following copyrights:
+ *    Copyright (C) 1999,2000  Frank Giannandrea
+ *    Copyright (C) 2000-2010  The Xastir Group
+ */
+
+
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif  // HAVE_CONFIG_H
+
+#include "snprintf.h"
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <unistd.h>
+#include <sys/stat.h>
+#include <ctype.h>
+#include <sys/types.h>
+#include <errno.h>
+
+// Needed for Solaris
+#ifdef HAVE_STRINGS_H
+#include <strings.h>
+#endif  // HAVE_STRINGS_H
+
+#include <Xm/XmAll.h>
+
+#ifdef HAVE_LIBCURL
+#include <curl/curl.h>
+#endif
+
+#include "xastir.h"
+#include "maps.h"	// only for MAX_FILENAME
+#include "main.h"
+
+#include "dlm.h"
+
+// Must be last include file
+#include "leak_detection.h"
+
+/**********************************************************
+ * DLM - DownLoad Manager - manage a list of items to download
+ **********************************************************/
+
+/* Notes on curl_multi and the tile queue:
+
+   We could have used a thread for each transfer, but that's messy and wasteful.
+
+   We could have used the curl_multi list as the queue, but that would end up
+   with us either waiting until the last few tiles were downloading before returning,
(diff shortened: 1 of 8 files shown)


tvrusso:
This looks like an interesting and exciting PR, and I will have to test it out a bit.  You might also advertise it on the Xastir mailing list and let some other folks test it out while it's pending.  I may not be able to get a look at it for a few weeks, and Curt is pretty busy, too.

It appears you may also have partially addressed issue report #23.  Have you tested this set of code changes against GraphicsMagick, in which GetIndexes is also deprecated?

ChocolateFrogsNuts:
I don't use GraphicsMagick, so it will need testing against that, but I only made the minimal changes to get a compile. The changes may well work with both.
Not much traffic on the mailing list recently but I'll announce it on there shortly anyway so we get some testers on board.

we7u:
Merged into my own local git repo, on SUSE Leap 42.3 with all updates and GraphicsMagick installed:

map_OSM.o: In function `draw_OSM_image':
/home/archer/xastir/src/map_OSM.c:726: undefined reference to `GetAuthenticPixels'
/home/archer/xastir/src/map_OSM.c:736: undefined reference to `GetAuthenticIndexQueue'
map_OSM.o: In function `draw_image':
/home/archer/xastir/src/map_OSM.c:525: undefined reference to `GetAuthenticPixels'
/home/archer/xastir/src/map_OSM.c:535: undefined reference to `GetAuthenticIndexQueue'
map_OSM.o: In function `draw_OSM_tiles':
/home/archer/xastir/src/map_OSM.c:1212: undefined reference to `SetImageBackgroundColor'
/home/archer/xastir/src/map_OSM.c:1242: undefined reference to `ClearMagickException'
/home/archer/xastir/src/map_OSM.c:1279: undefined reference to `IdentifyImage'
collect2: error: ld returned 1 exit status

we7u:
A few more things (all minor stuff):

- Changes to interface.c have nothing to do with the main reason for the PR. Would be better to do a separate PR for it so it's not hidden in the big patch.

- Changing from "<< 8" to "* 256"" is slower. I suggest keeping the "<< 8" code in the map*.c files.

- The headers on the two new files appear to have been copied from an existing file. They need Copyright dates updated at least. Should be only "Copyright (C) 2018 The Xastir Group" since they are new files. "dlm.h" says: "contributed by Jerry Dunmire, KA6HLD." which needs to be corrected as well. "dlm.c" has an $Id: message which is wrong, plus also has "This file derived from" message at the bottom of the header. Correct? Should be deleted?

I'm looking forward to trying this out when the GM stuff gets fixed!

ChocolateFrogsNuts:
Hi all,
Found the messages disappearing into my spam folder... and fixed that  :)


*interface.c:*
Yes that should probably have been a separate PR - I'm relatively new to
git so I'll see if I can manage to change that.


*<<8 vs *256:*
For int types the compiler will (at least should) optimize *256 into <<8
making it no slower.
The reason for the change is that the type of the variable changes to
double, so <<8 is no longer possible.

I may have to have another go at 3 printf calls that I changed (%i ->
%1.2f).  It's probably easiest to change them back to %i and cast the
args to match.


*Copyright:*
Yes, they were copied. Suggestions adopted.


*IM vs GM*
I'm no expert on IM or GM but I found deprecation info for IM 6.7 that
supports the function substitutions I made (compiling with 6.9 here).  
http://patrickdieudonne.com/ImageMagick-6.7.0-2/www/api/deprecate.html

GM has only deprecated GetPixels() and GetIndexes() from version 1.3.28
(Jan 2018), so that's not urgent.
I didn't find any info on what GM is replacing it with, but IM7 replaces
it yet again :(
 
For now I have modified the #if directives to only apply the changes
where GraphicsMagick is not in use.
That should get it to compile, and once we have a plan for how to handle
this I'll hunt down the remaining GetIndexes() calls that haven't been
replaced yet.


Mike


Curt Mills, WE7U wrote on 11/07/18 05:08:

ChocolateFrogsNuts:
Ok, I think I've taken care of everything requested, and I now have it
compiling on both ImageMagick 6.9.10 and
GraphicsMagick 1.3.30 without any related deprecation warnings.

I left the <<256 in because it's backward compatible, the compiler will
optimise it to <<8 for older
IM/GM and IM7 will require the change (6.9.10 already does and it's in
the documentation for porting to 7).


Give it a whirl
Mike


Curt Mills, WE7U wrote on 11/07/18 05:08:

tvrusso:
It would probably be good to get this resolved soon.  Curt:  have your requested changes been addressed to your satisfaction?

we7u:
Letting you know I'm looking at this again and have part of it checked. I'm really liking the code so far. I hope to have the check completed by this weekend.

tvrusso:
For what it's worth,  while I'm not reviewing this code, I have just downloaded the branch and am at least compile and run testing it on my system.  The graphicsmagick deprecation warnings are gone, and the code is running.  I am delighted with the new speed of the updates as I pan the map with an online map source, which used to be quite slow.

we7u:
Yep! You should probably read the comments at the top of dlm.c to see how he's doing it. Pretty impressive so far.

src/dlm.c
```diff
@@ -0,0 +1,926 @@
...
+
+/* DLM_queue_lock is held whenever reading or changing DLM_queue_entry or
+ * the linkages of the items in the queue.  Each item also has a lock
+ * for changing data in that item.
+ */
+xastir_mutex DLM_queue_lock = { .threadID=0, .lock=PTHREAD_MUTEX_INITIALIZER };
+pthread_t DLM_queue_thread;
+
+volatile int DLM_queue_progress_flag=0;
+volatile int DLM_queue_state=DLM_Q_STOP;
```
godfreja: Should DLM_queue_state be protected by a mutex to ensure correct memory visibility? I don't believe volatile is enough.

we7u: Are you ok if I merge this while we wait on the answer? I'm good with the rest of the code and think this would be a step forward. If the variable needs protection one of us can add that.

godfreja: I am OK with that, although it might be good to hold on a new release until we figure it out.

we7u: I'm confused as to how DLM_queue_state should be working. I could see it protected by a mutex and being an INT (for perhaps a count of download threads running), but I don't see how having multiple threads able to write specific values into it will work, mutex or not. It's a single variable but multiple threads can write STOP/RUN/QUIT/IDLE/STARTING values into it at any time. Perhaps make it a count variable with defined ALL-STOP value (-1) or something? It's also easily possible that I misunderstand what its purpose is.

ChocolateFrogsNuts:
One thread manages all the downloads, no matter how many downloads are
in parallel and that thread is a state machine which operates according
to DLM_queue_state.

The only time DLM_queue_state is changed outside of that thread is by
DLM_queue_start_if_needed(), and then ONLY when the downloader thread is
not running, thus no lock is required.

There may be some confusion coming from my re-use of DLM_Q_* for the
state of individual tiles as well as the state of the entire queue, but
that is fairly easy to spot as it always involves tile->state rather
than DLM_queue_state.


Good to see it getting some attention again :)

Mike
VK6MN


Curt Mills, WE7U wrote on 05/04/19 05:36:

we7u:
Excellent! That means we shouldn't have to do anything to the code. It's been merged with master. Thanks for your contribution!

src/dlm.c
godfreja: Just to clarify what my concern is: I've been bitten in the past where only one thread is modifying a variable and the reading thread does not see the update. The variable was marked volatile. A mutex is needed to ensure changes are visible to other threads. Having the mutex inserts a memory barrier.

See https://en.wikipedia.org/wiki/Memory_barrier#Example for a bit of an explanation.

godfreja: I can add in the mutexes if desired.

we7u: Feel free. Usual fork / modify(on a branch) / pull-request. I'd have to look deeper at the code to decide for sure that it's needed, but if one thread is writing and at least one other is reading - we need it. Thanks!

Issue #24, opened 2018-02-16
Building Xastir produces large numbers of warnings on newer compilers


tvrusso:
This was pointed out to me last week:  Xastir throws a very large number of compiler warnings on newer versions of GCC (specifically, GCC 6.x and newer).  It also throws a lot of warnings from Clang (which is my primary compiler).  Most of these are warnings about common errors that are not in fact real errors in Xastir code, but look like them (variables set but never used, variables possibly used uninitialized, but not ever actually used that way in Xastir, statements immediately after if statements that are indented as if they were intended to be guarded by the if, but which are in fact outside of the if, etc.).

I have started fixing these up, but wanted to record the issue here.

tvrusso:
To back up my claim of "a very large number of warnings,"  here are all the warnings emitted by GCC 6.3 on Raspian Stretch:

[pi_xastir_warnings.txt](https://github.com/Xastir/Xastir/files/1729939/pi_xastir_warnings.txt)

tvrusso:
This one is turning out to be really annoying.  It appears that the trend these days is to warn on just about everything that could possibly be a mistake, and not provide any means of shutting it up short of -Wno-\<warnaboutthis> or making crazy circumlocutions.

For example, certain functions like fgets and seteguid are set in their headers to be functions that should always warn if you don't use their results, e.g.:

      setegid(getgid());

Xastir has traditionally not warned or complained if this call fails, so it ignores the return value.  Some compilers (gcc 4.8.4 on Ubuntu 14.04) whine like mad about it --- but oddly gcc 6.3 on Raspbian Stretch doesn't.  One way to shut up the "unused result" warning is to assign the result to a variable and then don't do anything with it.  But then gcc 6.3 on Stretch complains that you've got a variable assigned but never used.  One can then make a nonsense use of the variable (e.g.:
    foobar=setegid(getgid());
    if (foobar != 0)
    {
      // do nothing
    } 

which does wonders for code readability.

fcc_data.c is *full* of uses of fgets that don't check their return values, and on Trusty with gcc 4.8 I get a ton of warnings.  Not on stretch with gcc 6.

Yay.

Anyhow, I've gotten all of the warnings in files from [a-i]*.c shut up on Stretch.  Not sure I should bother with the warnings from 4.8 on Trusty, especially all of those "unused result" warnings.  Then again, the steady stream of meaningless warnings does obscure the possibility of finding real errors, so it is probably worthwhile.

[...]

PR #57, opened 2019-04-08
Remove OpenTrac decoding routines

### LINKED ISSUE #56: Remove OpenTrac decoding routines
While working on issue #24, I find that interface.c's OpenTrac support code spews quite a few warnings about strncat misuse.  Investigating, some of these actually appear to point to the possibility of buffer overrun.  This could be resolved by making sure that the code correctly tests string lengths of OpenTrac packets and uses safer functions than strncat.

However, OpenTrac is essentially a dead protocol --- it was proposed as something of a replacement for the APRS protocol, but never caught on.  It was pretty much only supported by a few devices, all by the same creator as the one who proposed OpenTrac itself.  

We do not believe anyone is using this code anymore, and rather than expend energy maintaining it, it would be best to remove it.  Code removed is code that won't rot.

tvrusso:
OpenTrac was a hot new thing back around 2003.  It is mostly dead now.

And worse, the OpenTrac code is now generating a lot of warnings from
GCC 8, some of which look like they might actually point to potential
errors.  Since the code may be borken, and nobody uses the OpenTrac
protocol anymore, this commit simply removes the functions that handle
decoding those packets in the interests of making the cleanup of
warnings (issue #24) easier.

Resolves #56.


we7u:
I'm good with removing it from the master branch as long as we have a way to incorporate it back in later somewhat easily. Of course we'll have to fix the compiler warnings later in that case.

tvrusso:
Reverting the single commit in the PR will bring everything back.  Nothing gets lost permanently so long as it's in git.

PR #58, opened 2019-04-11
Add support for GNGGA and GNRMC sentences

### LINKED ISSUE #53: Unable to get GPS data from DRAWS hat [because Xastir does not support GNGGA or GNRMC sentences]
I've complied the most recent xastir build (2.1.1). I've enable Networked GPS (via GPSD) and the interface show UP. However I'm not getting the GPS data to read and update my position. I've tested using YAAC and I am getting updated GPS data there via GPSD. I have verified that GPSD is running

gpsd.service - GPS (Global Positioning System) Daemon
   Loaded: loaded (/lib/systemd/system/gpsd.service; enabled; vendor preset: ena
   Active: active (running) since Thu 2019-04-04 13:49:10 PDT; 30min ago
  Process: 500 ExecStart=/usr/local/sbin/gpsd $GPSD_OPTIONS $OPTIONS $DEVICES (c
 Main PID: 503 (gpsd)
   CGroup: /system.slice/gpsd.service
           └─503 /usr/local/sbin/gpsd -n /dev/ttySC0 /dev/pps0

I am running on a RBI 3B+ and using a DRAWS hat for my GPS source.

tvrusso:
GLONASS receivers send $GNGGA sentences instead of $GPGGA sentences, and
$GNRMC instead of $GPRMC.  But the only difference between these
sentences is the GN instead of GP --- the data is identical.

This appears to be the reason that users of the DRAWS hat are unable
to use this device with Xastir, even though it appears to work with
all other programs that consume NMEA sentences.

This commit attempts to recognize GGA and RMC sentences irrespective
of whether they begin with "$GP" or "$GN".  It is tested so far only
with GPS receivers, and it doesn't break those.  I'm adding this on a
branch so that users of the DRAWS hat can test it out before we
actually merge into master.

If it works, this closes #53

### DIFF
--- a/src/gps.c
+++ b/src/gps.c
@@ -138,13 +138,13 @@ int decode_gps_rmc( char *data,
     if ( (data == NULL) || (strlen(data) < 37) )
         return(0);  // Not enough data to parse position from.
 
-    if (strncmp(data,"$GPRMC,",7) != 0)   // No GPRMC found
+    if ((strncmp(data,"$GPRMC,",7) != 0)&&(strncmp(data,"$GNRMC,",7) != 0))   // No GxRMC found
         return(0);
 
     if(strchr(data,',') == NULL)  // No comma found
         return(0);
 
-    (void)strtok(data,",");   // get GPRMC and skip it
+    (void)strtok(data,",");   // get GxRMC and skip it
     temp_ptr=strtok(NULL,",");   // get time
 
     if (temp_ptr == NULL)   // No comma found
@@ -362,6 +362,10 @@ int decode_gps_rmc( char *data,
 //
 // $GPGGA,170834,4124.8963,N,08151.6838,W,1,05,1.5,280.2,M,-34.0,M,,,*75 
 // $GPGGA,104438.833,4301.1439,N,08803.0338,W,1,05,1.8,185.8,M,-34.2,M,0.0,0000*40
+
+// NOTE:  GLONASS receivers can output GNGGA strings that are identical in
+// form to GPGGA, but with the second character being different (N instead of P)
+// Let's try to support those, too.
 //
 int decode_gps_gga( char *data,
                     char *long_pos,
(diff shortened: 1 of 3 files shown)


tvrusso:
We have a report on issue #53 that the changes in this PR do indeed get the DRAWS hat working, so I'm changing it from a draft PR to a real one, and making it ready for review.

we7u:
You may want to assure support for Beidou and Galileo too. I have some GPS/Glonass chips at home right now that would work with your mods, but am planning to play with a choose-3-out-of-4 chip soon that has GPS/Glonass/Beidou/Galileo in it. Not sure if there is a 5th constellation yet.

tvrusso:
I think that's a reasonable suggestion, but requires a bit more thought than simply slamming in a second pair of "strncmp" calls as I have done here.  What I think is needed is for those lines to recognize "$G" as the first two characters, and "GGA," or "RMC," as the fourth through 7th chars, instead of recognizing two distinct 7-character starting strings.  Jason (KG4WSV) had suggested that we'd want to recognize "G[PLABN]" here rather than just GP, and all I've done is make it recognize GN or GP. 

Lemme study on it, and maybe I'll add commits to this branch.  Unfortunately, I have no means of testing anything other than GP strings, as I don't have any receivers for these other constellations.

tvrusso:
K.  Latest commit on this branch should extend our support to GLONASS, GNSS, Beidou, and Gallileo.

Note that what I had previously called "GLONASS" support was actually GNSS support.  The sentences are:
   $GA --- Gallileo
   $GL --- GLONASS
   $GB --- Beidou
   $GN --- GNSS
   $GP --- GPS

I supported this by creating two new functions, isGGA and isRMC, then replacing all of the code that tried to do straight string compares to specific strings with calls to these two.  Then this function checks:
   - Does the string start with $G?
   - Is it more than 7 characters long?  And are characters 4-7 "GGA," (or "RMC,")
   - and finally, is the third character one of P,N,L,A, or B.

This should be more extensible in the future if more geolocation constellations get launched and receivers for them start showing up in the APRS world.

tvrusso:
I just force-pushed a rebased branch for this PR, after having merged #60.

If you are testing out the branch, go ahead and switch back to master, delete the branch (git branch -D feature-supportglonass) and check it out again.

PR #61, opened 2019-04-15
Removing the internal Shapelib tree

we7u:
Removing the stale version of Shapelib that we compiled in if Shapelib
was not found on the system. Most systems have Shapelib available now
or can obtain it easily, so keeping an old copy of Shapelib in the
Xastir sources was deemed unworkable. This should also fix the
"distclean" issue that Tom Russo explained on the Xastir list on
04/15/2019: I encountered that bug personally while trying to build
inside a "build" directory.


tvrusso:
Beautimous.  I'd be happy to see this merged.

configure.ac
```diff
@@ -402,20 +391,16 @@ if test "${shapelib_desired}" = "yes" -a "${force_internal_shapelib}" = "no"; th
...
  AC_MSG_WARN([**************************************************************** ])  
  LIBSHP="-Lshapelib -lshape"
  CPPFLAGS="-I\$(top_srcdir)/src/shapelib $CPPFLAGS"
- SUBSHP="shapelib" 
- use_shapelib=yes
- force_internal_shapelib=yes
+ use_shapelib=no
  AC_DEFINE(HAVE_DBFGETFIELDINDEX, , Define to 1 if your `shp' library has DBFGetFieldIndex. ) 
  AC_DEFINE(HAVE_LIBSHP, , Define to 1 if you have the `shp' library (-lshp). )
  AC_DEFINE(HAVE_SHAPEFIL_H, , Define to 1 if you have the `shapefil.h' header. )
```
tvrusso: I believe this is in error.

It retains the setting of HAVE_LIBSHP, HAVE_SHAPEFIL_H, and HAVE_DBFGETFIELDINDEX even while saying we don't have shapelib at all.

we7u:
FWIW: I tried it with and w/o the shapelib and shapelib-dev packages installed on this machine. It doesn't compile w/o them in there, and it works when they are in there. Good to go? If so I'll merge.

configure.ac
tvrusso: That is, if we're going to allow configure to proceed with a notice to install shapelib to silence a notification, we should also make the build work.  And it won't as it is.

I think if you remove the three AC_DEFINEs, it'll be right.

tvrusso: Actually, I think you have to remove everything other than the warning here --- including the setting of LIBSHP and CPPFLAGS.   Only the warning text and the setting of use_shapelib=no should be inside the conditional there.  Then it should once again be possible to build Xastir without shapelib support (and have it really crippled).

If you want it to be a *mandatory* TPL, then configure should emit an *ERROR* about shapelib being unavailable, and not leave it to compile time to fail.

tvrusso: More comments.  Sorry. 

If you look at the LIBSHP variable, its entire purpose is to handle passing the internal shapelib -l/-L flags around.  So it should be removed from configure.ac (along with the AC_SUBST commands that pass it down to Makefile.ams), and should also be removed from src/Makefile.am in two places (where it appears as "@LIBSHP@"

we7u:
Am getting this at the link stage when Shapelib packages installed (no errors when not installed):
/usr/lib64/gcc/x86_64-suse-linux/4.8/../../../../x86_64-suse-linux/bin/ld: testdbfawk.o: in function `main':
/home/archer/src/we7u/xastir/github-we7u/Xastir/src/testdbfawk.c:202: undefined reference to `dbfawk_sig'
/usr/lib64/gcc/x86_64-suse-linux/4.8/../../../../x86_64-suse-linux/bin/ld: /home/archer/src/we7u/xastir/github-we7u/Xastir/src/testdbfawk.c:209: undefined reference to `dbfawk_find_sig'
/usr/lib64/gcc/x86_64-suse-linux/4.8/../../../../x86_64-suse-linux/bin/ld: /home/archer/src/we7u/xastir/github-we7u/Xastir/src/testdbfawk.c:226: undefined reference to `dbfawk_field_list'
/usr/lib64/gcc/x86_64-suse-linux/4.8/../../../../x86_64-suse-linux/bin/ld: /home/archer/src/we7u/xastir/github-we7u/Xastir/src/testdbfawk.c:229: undefined reference to `dbfawk_parse_record'
/usr/lib64/gcc/x86_64-suse-linux/4.8/../../../../x86_64-suse-linux/bin/ld: /home/archer/src/we7u/xastir/github-we7u/Xastir/src/testdbfawk.c:263: undefined reference to `dbfawk_free_sigs'
/usr/lib64/gcc/x86_64-suse-linux/4.8/../../../../x86_64-suse-linux/bin/ld: /home/archer/src/we7u/xastir/github-we7u/Xastir/src/testdbfawk.c:160: undefined reference to `dbfawk_load_sigs'

tvrusso:
Well, that does seem to be another problem.  Is that with the unmodified code in this PR, or with further changes?

we7u:
Further changes: Removing the three AC_DEFINE's at the tail end and modifying the src/Makefile. Let me push that up in the pull request. I'm not good with configure.ac stuff...

we7u:
I added one of the AC_DEFINE's back and it didn't fix it. Will push shortly.

tvrusso:
BTW, I can't actually test out this set of changes right now because I'm at work --- I was just reviewing the PR by looking at the changes.  Please hold off on merging it until we get this sorted out.  I can pull down your changes tonight and see if I can't figure out what's wrong with them.  

dbfawk_sig and its friends are defined in dbfawk.c, which is conditioned on HAVE_LIBSHP, HAVE_LIBPCRE, and WITH_DBFAWK.  Sounds like somthing is borken in the setting of one or more of these downstream of your changes.

configure.ac
```diff
@@ -386,22 +385,16 @@ if test "${shapelib_desired}" = "yes" ; then
     LIBS="$LIBS -lshp"
     AC_CHECK_LIB(shp,DBFGetFieldIndex,
       AC_DEFINE(HAVE_DBFGETFIELDINDEX, , Define to 1 if your `shp' library has DBFGetFieldIndex. ) )
-    AC_DEFINE(HAVE_LIBSHP, , Define to 1 if you have the `shp' library (-lshp). )
```
tvrusso: HEEEEEERE's your problem.  You should not remove this AC_DEFINE, because it is necessary to enable compilation of various ifdef'd code in dbfawk.c and map_shp.c.

configure.ac
```diff
@@ -386,22 +385,16 @@ if test "${shapelib_desired}" = "yes" ; then
...
  AC_MSG_WARN([Your system does not have shapelib installed.                    ])
  AC_MSG_WARN([Install shapelib on your system to eliminate this warning.       ])
  AC_MSG_WARN([**************************************************************** ])  
- LIBSHP="-Lshapelib -lshape"
- CPPFLAGS="-I\$(top_srcdir)/src/shapelib $CPPFLAGS"
  use_shapelib=no
- AC_DEFINE(HAVE_DBFGETFIELDINDEX, , Define to 1 if your `shp' library has DBFGetFieldIndex. ) 
- AC_DEFINE(HAVE_LIBSHP, , Define to 1 if you have the `shp' library (-lshp). )
- AC_DEFINE(HAVE_SHAPEFIL_H, , Define to 1 if you have the `shapefil.h' header. )
+ AC_DEFINE(HAVE_DBFGETFIELDINDEX, , Define to 1 if your `shp' library has DBFGetFieldIndex. )
```
tvrusso: THIS AC_DEFINE should be removed.

src/Makefile.am
```diff
@@ -92,9 +92,9 @@ compiledate.c: $(XASTIR_SRC)
 remove_compiledate:
 	rm -f compiledate.c compiledate.o
 
-xastir_LDADD=@LIBRTREE@ @LIBSHP@
+xastir_LDADD=@LIBRTREE@
 
-testdbfawk_LDADD=@LIBSHP@
+testdbfawk_LDADD=
```
tvrusso: You should probably jjust remove this line, since it is not doing anything.

tvrusso:
What happened is you removed one too many AC_DEFINEs.  You removed *ALL* the HAVE_LIBSHP AC_DEFINE lines, not just the one inside the block that should only execute when shapelib doesn't exist.  So when shapelib DOES exist, the macro wasn't getting set.

tvrusso:
What I'd been talking about in the earlier comment about removing the "LIBSHP" variable was not referring to the HAVE_LIBSHP tests.

I'll go ahead and make a line-by-line commentary in the changes, so it's really more clear what I'm talking about.

tvrusso:
Actually, as it stands, I don't need to add any more commentary than is there.  You removed the correct lines for the LIBSHP variable as I intended.  It's just that you also did the overzealous removal of the AC_DEFINE(WITH_LIBSHP in that one case.

we7u:
Compiles for both cases now. Thanks!

tvrusso:
Will look it over again more carefully tonight and resolve the review.

we7u:
Verified that switching to a "build" directory and getting the "must run distclean" thing works now. I went up one directory, ran make distclean, back into build, and was able to compile.

tvrusso:
Did a few different builds with and without shapelib and it looks good.  Ready to be merged, I think.

we7u:
Did a pull into my laptop's github clone. It still ended up with the src/shapelib/ directory and files, but they were then untracked. Any way to get Git to delete the files on the local clone when I do a pull? I already did an "rm -rf shapelib" so they're gone... Just thinking about others that might end up with these files hanging around.

tvrusso:
Looks like the shapelib directory and its "contrib" are not removed immediately by the git pull because they contain bootstrap artifacts (Makefile.in) that were marked as "ignored" in .gitignore.  In my primary desktop clone, everything's gone *but* those artifacts.  They don't even show up as untracked, ignored files (git status -u --ignored).  

The actual shapelib source files are in fact removed, as are the contents of contrib.

On my laptop, I did a "git clean -dXf" to get rid of bootstrap artifacts first, then did a git pull.  The shapelib and shapelib/contrib directories were deleted when I did the git pull.

we7u:
Fun! Got to another machine and did a "git fetch" and then later a "git pull". In this clone I compiled w/o the "build" directory previously and in fact was using both methods (with and w/o "build") to test things. I still see "shapelib" and "shapelib/contrib" directories because "Makefile.in" and "Makefile" files are present in each. This is probably ".gitignore" rearing its head again 'cuz of a repo-wide regex. I also see ".deps" directories in each one with "*.Po" files in them. These remnants are probably not worth worrying about as people can just "rm -rf shapelib" and be done with them. Perhaps a note to the list would be wise but that's about it. Over time as people check out new clones the problem will go away on its own.

tvrusso:
Yeah, doing the "git clean -dXf" (removing ignored files and untracked directories) prior to the pull is pretty much required in this case, or untracked and ignored files will remain in the directories that got deleted.

A "git clean -dXf" after the pull would probably also clean it out.

tvrusso:
Actually, a "git clean -Xf" to get rid of the ignored files (Makefile.in and Makefile) followed by a "git clean -df" (to get rid of untracked directories) is required.  Just tested this on a third, seldom used clone I had on my igate machine.

The issue here is that the ignored files prevent git from removing directories, so the "git -Xf" takes care of those.  Then you have to get rid of the untracked (but not ignored) empty directories with the second command.

Using git commands to get rid of them is a little safer than just using rm (unless you know exactly what files to delete).

we7u:
Well I could delete that fork, but I have other branches there (feature-greeklanguage), so won't. I could delete the branch though.
I'm also not going to run that "git clean" command as I have untracked files that I don't want to lose. "rm -rf" on the specific directory works for me.

tvrusso:
It should never be necessary to delete a clone just to clean it up.  It is better to learn the right git commands to clean it up.
Managing untracked stuff isn't hard with git, either.  You can always do a "git stash --include-untracked", which pushes the current state of the working directory (including untracked files) into the stash, then cleans up untracked files.  You can then do the other cleaning operations, followed by a "git stash pop" to pull the stash contents back into your working directory.

PR #67, opened 2019-04-21
Ensure x_spider children don't return

### LINKED ISSUE #65: Duplicate xastir windows appeared
While testing some code changes I saw multiple xastir windows appear. The stdout output indicated that one copy started and at a certain point multiple copies were competing for network ports. I ultimately fixed the problem by killing some old xastir processes that were running even though there were no xastir windows.

I suspect something isn't handled well (in the network code?) where a fork happens and if there is an error (due to say something else having a port open) the forked copy returns to the "normal" control flow instead of behaving as a proper child and two copies show on the screen. This is pure speculation.

This also raises the question of why I had some old xastir processes hanging around with no windows open.

godfreja:
If the UDP x_spider server couldn't open any ports (such as due to
another xastir process having them open) the child would return to
the main control logic of xastir, leading to two copies of xastir
running and duplicate windows appearing. Add an exit(1) after the
call to UDP_Server() so this can't happen. This fixes #65.

The TCP server didn't use return, but I added an exit(1) after the
call to TCP_Server() just to be safe.

I also fixed some warnings on MacOS compiles. This helps with
issue #24.

### DIFF
--- a/src/x_spider.c
+++ b/src/x_spider.c
@@ -833,8 +833,11 @@ extern char *__progname, *__progname_full;
 #endif  // __linux__
 static char *LastArgv = ((void *)0);
 static char **local_environ;
+#ifdef __linux__
+#ifndef __LSB__
 static char *old_progname, *old_progname_full;
-
+#endif  // __LSB__
+#endif  // __linux__
 
 
 void clear_proc_title(void)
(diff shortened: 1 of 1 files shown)


tvrusso:
I had never before observed the double-window error of bug #65, but today I tried to make it happen by firing up Xastir (without this fix) on my Ubuntu box, enabling server ports, then deliberately killing the main Xastir process --- leaving the two server port processes as zombies.  When I fired up Xastir again, sure enough, it spewed error messages and put up two windows.

The same thing did NOT happen on my primary desktop machine, which spewed the errors but did not create a second window.  That is probably a window manager thing --- I run fvwm2 on the primary machine, and gnome desktop on the Ubuntu machine.

After pulling the fix branch to my laptop, the double-window thing no longer happens.

PR #68, opened 2019-04-21
Fixing compiler warnings

we7u:
Looks like I fixed compiler warnings for the three Linux Travis-CI builds but not the OSX build. It's a start!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ script:
 matrix:
   include:
     - os: linux
-      name: "Linux full featured with GraphicsMagick"
+      name: "Ubuntu 14.04 full-featured with GraphicsMagick"
       addons:
         apt:
           packages:
@@ -30,7 +30,7 @@ matrix:
             - gpsmanshp
     - os: linux
       dist: xenial
-      name: "Ubuntu 16.04 full featured with GraphicsMagick"
+      name: "Ubuntu 16.04 full-featured with GraphicsMagick"
       addons:
         apt:
           packages:
@@ -50,7 +50,7 @@ matrix:
             - gpsman
             - gpsmanshp
     - os: linux
-      name: "Linux light with ImageMagick"
+      name: "Ubuntu 14.04 light with ImageMagick"
       addons:
         apt:
           packages:
@@ -63,7 +63,7 @@ matrix:
             - libshp-dev
             - libmagickcore-dev
     - os: osx
-      name: "MacOS"
+      name: "OSX"
       addons:
         homebrew:
           packages:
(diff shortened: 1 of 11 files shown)


tvrusso:
The Mac OS X warnings in wx.c could easily be fixed by addressing the issue of comparing the pointer to a statically declared array to zero, which is always false and therefore pointless:

if ( (computed_gust > 0.0) || (weather->wx_gust != 0) )
{
   (print into wx_gust)
}

Because the wx_gust array is declared as     char    wx_gust[4] in database.h, this conditional means that the gust is ALWAYS being printed into wx_gust.

Removing the unnecessary compare would silence the remaining wx.c warnings.

tvrusso:
One more comment on the Mac OS X warnings in wx.c:  I just sifted through the history of database.h (which used to be called "db.h"), and wx_gust has been declared as "char wx_gust[4]" since the day the file was imported into CVS in 2002, as was at least one of the conditionals "if ( (computed_gust >0.0) || weather->wx_gust !=0)" --- meaning that since the dawn of time, this conditional has *always* been an unconditional.

So the right thing to do to avoid changing behavior is to remove the conditional altogether, which will silence the warning *AND* make sure we do what we have always done.

we7u:
Regarding wx_gust[]: Would that really be the correct way to fix it? Seems to me the intent was an && there instead of an ||. According to "git blame" it was either Chuck Byam or myself on those lines of code. I believe the intent was to make sure we weren't trying to write into a null pointer, so removing the second part of the condition looks like the correct fix. I'll do that.

we7u:
Updated. Ok to merge so we can get on to the next group (and probably fix some OSX ones later)?

tvrusso:
I was not sure that making it unconditional was the right fix, but it _is_ the thing that would make sure that no change in behavior would result.

I agree, it looks like it should have been the intent to use "&&" here instead of ||, in which case simply removing the condition restores it to the intended function.  However, if you go all the way back to the very first commit of wx.c (which was when it was imported into SF CVS in 2002, I think), there was already one line with the OR version in it --- the others look to have been cut/pasted from that original mistake.  That is, while it *may* have been the intent for it to behave this way, it NEVER has.

The one concern I had about simply removing the nugatory condition is that it suddenly makes it possible that computed_gust is zero, and then NOTHING gets written into the wx_gust string -- i.e. it'll wind up with old data.  

Since 2002, and probably long before, Xastir has always written the value of computed_gust into the string, whether it's zero or not.  So removing the condition *could* cause stale data to remain in there if we compute the gust to be 0.0.  I see nothing that would clear wx_gust out.

I also do not have any weather station that would exercise the computed_gust code, because the LaCrosse/Davis METEO code reads the gust from the device and doesn't compute it.

So, if you are comfortable correcting the logic error and changing the behavior, I'm OK with it, but you should make sure it's really what you want to do.

tvrusso:
If you are sure you want to change that wx_gust code the way you have, I have no other comments.  Merge the shiznit at your discretion.

we7u:
Now that you pointed out the zero issue I'm going to agree with you and remove the conditionals altogether.

PR #70, opened 2019-04-22
Warning fixes 2

we7u:
Partial fix for bug #24. Fix compiler warnings plus get rid of "register" keyword where used.

### DIFF
--- a/src/map_WMS.c
+++ b/src/map_WMS.c
@@ -159,7 +159,7 @@ void draw_WMS_map (Widget w,
     register long map_c_T, map_c_L; // map delta NW edge coordinates, DNN: these should be signed
     register long tp_c_dx, tp_c_dy; // tiepoint coordinate differences
     unsigned long c_x_min,  c_y_min;// top left coordinates of map inside screen
-    unsigned long c_y_max;          // bottom right coordinates of map inside screen
+//  unsigned long c_y_max;          // bottom right coordinates of map inside screen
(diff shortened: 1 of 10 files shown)


tvrusso:
This looks good to me.

I'm still seeing some warnings on OS X in files that were touched by this commit and would like to see them fixed, ultimately, but that can happen in future hackage.

The ones that concern me are the ones about "abs( a - b)" where a and b are both unsigned longs --- OS X is warning that because they're both unsigned long, by C type rules, so is a-b, and therefore the abs can have no effect --- that is, even though a could be smaller than b and in real world math could be negative, in unsigned math it's actually never going to be negative --- just wrong.

That is one that needs more thought.  It is probably also wrong on other systems, because it goes to the heart of C implicit typing rules for expressions, but they just don't warn about it.

I am also concerned about the warnings in the Ubuntu 14.04 ImageMagick build of map_geo.c, because they're mentioning deprecated function calls that I thought we got fixed as part of bug #23 and PR #40.

I'm going to go ahead and merge this PR, because it does indeed improve the warning footprint complained about in #24.

tvrusso:
I would, however, be much happier if your commit messages (especially their titles) were a little more descriptive.  A quick "git log" without "-p" would be a difficult read if one were searching for specific explanations of what was addressed and why in each commit.  Sure, the PR on github does help (so long as it's appropriately mentioned in the commit message), but the messages should be more informative on their own.

PR #71, opened 2019-04-22
Reorder bounds checking for while loop in draw_shapefile_map

godfreja:
I've started playing with using clang's "-fsanitize=address" feature
to look for bugs. When shapefile maps were enabled it crashed on a
read buffer overlow. We were reading the value from a char array
and then checking to see if the index was in bounds. Reverse it
so we check for in bounds first.


tvrusso:
Fascinating.  This error was introduced in commit c926df8a in Feb of 2002.

PR #75, opened 2019-04-23
Fix buffer overflow in jump_sort()

godfreja:
We were passing 200 into sort_input_database as the size of name,
which was only 100 in size. Switch the two hardcoded array sizes
in jump_sort() to use sizeof().

### DIFF
--- a/src/location_gui.c
+++ b/src/location_gui.c
@@ -181,7 +181,7 @@ void jump_sort(void) {
     f=fopen(location_file_path,"r");
     if (f!=NULL) {
         while (!feof(f)) {
-            (void)get_line(f,temp,200);
+            (void)get_line(f,temp,sizeof(temp));
             if (!feof(f) && strlen(temp)>8) {
                 temp_ptr=strtok(temp,"|");  /* get the name */
                 if (temp_ptr!=NULL) {
@@ -189,7 +189,7 @@ void jump_sort(void) {
                         sizeof(name),
                         "%s",
                         temp);
-                    (void)sort_input_database(location_db_file_path,name,200);
+                    (void)sort_input_database(location_db_file_path,name,sizeof(name));
                 }
             }
         }


tvrusso:
This looks like the correct fix to a mistake that has been in the code since before the beginning of the CVS/Git history we still have (meaning it was present in the CVS repo from which sourceforge's CVS repo was initialized, with all history lost).  See commit fa09e55.

godfreja:
Taking another look at sort_input_database I'm concerned changing the size passed in is going to mess up existing locations_db.dat-ptr files. This code:

            (void) fseek(my_data,0l,SEEK_END);  //KD6ZWR

            new_data_ptr = data_ptr = ftell(my_data);
            (void)fwrite(fill,(size_t)size,1,my_data);
            records = (data_ptr/size);

calculates the number of records in the file based on the size passed in. The number of records it considers previously written files to have will now double. 

I wonder if name in jump_sort() should be sized to 200 to correct for this. I didn't see any issues when I tested my change, but I'm still a bit nervous.

I won't be able to look at this more until tomorrow evening. If anyone has any thoughts on this please comment.

godfreja:
The code deletes the database files before calling jump_sort so the change in record size should be fine.

tvrusso:
It was a mistake to merge this PR, as noted in #93.  Changing the one instance of "200" to "sizeof(name)" in this one location actually caused the database to be written out in chunks of 100 bytes instead of 200, but scattered all over the rest of the file are accesses of the file that do it 200 bytes at a time.  The number of magic 200s all over is actually a little scary.

The result of changing the 200 to sizeof(name) is that map bookmarks that are present in ~/.xastir/config/locations.sys don't all show up in the map display bookmarks dialog, presumably because the database is being read in as if it were 200 byte records, but each record is actually 100 bytes.

I've reverted the commit and will have to investigate what the correct fix is.

PR #77, opened 2019-04-23
Remove "misleading indentation" warnings from two files

tvrusso:
This is a WHITESPACE-ONLY pull request.  It removes a handful of "misleading indentation" warnings when building with GCC 6 or GCC 8.  A misleading indentation warning is emitted when this sort of thing is found:

   if (foo)
       bar;
       baz;

This is indented as if baz is inside the if, but it is not guarded by the if.  

I did this by reindenting map_geo.c and map_tif.c according to project standards, which not only removed the warnings but also fixed some other abuses of indentation style.

No code changes or comment deletions were made in this series of commits.  It took great restraint not to blow away (or even re-indent) blocks of commented-out code.  But I did in fact exercise that restraint and change nothing but white space.

Addresses more warnings per bug #24.

### DIFF
--- a/src/map_geo.c
+++ b/src/map_geo.c
@@ -161,12 +161,12 @@
 
 int check_interrupt( 
 #ifdef HAVE_MAGICK
-			Image *image, ImageInfo *image_info, ExceptionInfo *exception, 
+                    Image *image, ImageInfo *image_info, ExceptionInfo *exception, 
 #else // HAVE_MAGICK
-			XImage *xi,
+                    XImage *xi,
 #endif // HAVE_MAGICK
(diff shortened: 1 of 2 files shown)


tvrusso:
Note that these warnings *ONLY* show up in later versions of gcc, not in anything before gcc 6.  GCC 6 is used in Debian Stretch (also Raspbian Stretch), gcc 8 in FreeBSD.  Ubuntu 16.04 and 14.04 (the two we test on with Travis) don't have a recent enough version of gcc to show the warnings, and would have to have non-standard PPAs enabled to get anything newer installed.

PR #86, opened 2019-04-23
Fixing a print format string

we7u:
Small patch but gets rid of another compiler warning. Part of fixes for bug report #24. Please check this one carefully to make sure it's the correct way to fix it.


tvrusso:
Why is this code even using snprintf at all?  The entire purpose of it is to make "printer_program" the null string.  Why not replace the call to xastir_snprintf with "printer_program[0]='\0';" and save a function call?

we7u:
Feel free to deny this PR and code it the other way.

we7u:
Oops. Wrong button.

we7u:
I'll do another PR on it shortly.

PR #87, opened 2019-04-23
Libax25

we7u:
Fixing another compiler warning (variable set but not used) that shows up on the minimal Ubuntu-14.04 build when libax25 is not present. This is part of the fixes for issue #24.

### DIFF
--- a/src/interface_gui.c
+++ b/src/interface_gui.c
@@ -4807,8 +4807,8 @@ void Config_AX25( /*@unused@*/ Widget w, int config_type, int port) {
 #else
         // Need code to use button_ok variable to quiet a compiler warning
         //when we don't have LIBAX25 linked-in.
-        if (button_ok != null) {
-            // Do nothing
+        if (button_ok == button_ok) {
+            // Do nothing (to shut up a compiler warning)
         }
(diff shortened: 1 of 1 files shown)


tvrusso:
On second thought, this PR seems to be a combination of an alternate fix for closed PR #86 and a separate issue, but the fix of the null string has a merge commit log instead of a descriptive one.

I can fix it up manually and do the merge myself.  Will keep this PR open for a bit while I do that.

tvrusso:
I rebased and cleaned up the branch myself and pushed.

tvrusso:
Actually, the method chosen to "do something" here introduces a new "tautological compare" warning from Clang, since "button_ok == button_ok" is always true.  I'll fix that up.

tvrusso:
I changed it to "button_ok == button_cancel" (which will always be false at run time, but the compiler can't tell), so now it should get neither the unused variable NOR the tautological compare warning.

we7u:
Tom: I'm uncomfortable at this point with _any_ rebasing, so I'm glad you took care of it. Will ask question on the -dev list somewhat related to this.

PR #95, opened 2019-04-25
Dlm threading

we7u:
This should fix any GUI threading issues in dlm.c. I commented out the Xm function call, the write to the status line, the variable being used for the string and the xastir_snprintf() that set the variable.

Once this has undergone a bit of testing I can go in and remove those bits altogether. That can be the last step of the bug report before it is closed (assuming this attempt is successful).

### DIFF
--- a/src/dlm.c
+++ b/src/dlm.c
@@ -44,7 +44,7 @@
 #include <strings.h>
 #endif  // HAVE_STRINGS_H
 
-#include <Xm/XmAll.h>
+//#include <Xm/XmAll.h>
 
 #ifdef HAVE_LIBCURL
 #include <curl/curl.h>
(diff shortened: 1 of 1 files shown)


tvrusso:
Sure, this will take out all use of X from the DLM, and should eliminate any possibility  of thread-unsafe usage.  It does, unfortunately, mean that a user will have no indication of what the DLM is doing as it is doing it, but perhaps that is a sacrifice that we can live with.

I am unable to do any testing on this at all, because I have never seen the crash despite many, many tries on three different OSen, and further I am unavailable for all Xastir work for the next few days.  Someone who actually can make the thing fail will have to review it to see if it actually fixes the map crashes of #66.

godfreja:
This code change looks fine to me. I can't reliably reproduce the error this is trying to address, so I can't test if this is an actual fix. It looks like it is.

Issue #72, opened 2019-04-22
Potential problem with mixture of long/unsigned long/abs in map_tif.c


tvrusso:
map_tif.c contains some lines of code that cause clang on OS X (and clang in general on other OSen) to throw some warnings about use of abs() on an unsigned expression.

Investigating, this is not just a warning that should be removed while working on #24, but an actual usage error that probably needs fixing correctly, not just warning silence effort.  It could (but does not in any case I've experimented with) produce wrong answers in addition to warnings.

A typical line in map_tif.c that causes this warning is:
```
        xastir_left_x_increment = (float)
            (1.0 * abs(SW_x_bounding_wgs84 - NW_x_bounding_wgs84)   // Need to add one pixel worth here yet
            / abs(SW_y - NW_y));
```
In this code, xastir_left_x_increment is of type float, SW_x_bounding_wgs84 and NW_x_bounding_wgs84 are both unsigned longs, and SW_x and NW_x are both signed ints.

The warning is this:

```
map_tif.c:1999:20: warning: taking the absolute value of unsigned type 'unsigned long' has no effect [-Wabsolute-value]
            (1.0 * abs(SW_x_bounding_wgs84 - NW_x_bounding_wgs84)   // Need to add one pixel worth here yet
                   ^
map_tif.c:1999:20: note: remove the call to 'abs' since unsigned values cannot be negative
            (1.0 * abs(SW_x_bounding_wgs84 - NW_x_bounding_wgs84)   // Need to add one pixel worth here yet
                   ^~~

```
The problem here is that according to standard C expression typing rules, the difference between two unsigned ints is itself an unsigned int --- so even if SW_x_bounding_wgs84 is smaller than NW_x_bounding_wgs84, the difference between them is NEVER negative --- it's an unsigned value, and is carried out in unsigned arithmetic.  abs() is supposed to take a signed int and return a signed int

This code has existed in this way in map_tif.c since map_tif.c was split out from maps.c in commit 2bcc1e9, and existed in maps.c since the beginning of history in commit fa09e5594.

But it's  technically wrong (according to C type conversion rules) whenever the difference in the numerator's abs call is negative, but **_clang's suggestion to remove the call  (which does indeed silence the warning) actually causes it to produce wrong answers_**, which is even worse.

[...]

PR #104, opened 2019-04-27
Osx warnings

we7u:
Implemented a bunch of fixes for compiler warnings that were showing up in the OSX build. This is part of the fix for bug #24.

### DIFF
--- a/src/map_OSM.c
+++ b/src/map_OSM.c
@@ -195,7 +195,7 @@ static long osm_scale_x(long xastir_scale_x) {
        if (xastir_scale_x > osm_level[i]) {
             continue;
         } else {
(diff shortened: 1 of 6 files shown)


tvrusso:
Per #72, I believe there are some problems with commit e51b371 and commit 6ff17d6.

In commit 6ff17d6, the difference between two unsigneds is cast to an int.  This almost certainly works fine, but what one really wants is to cast both unsigneds to an int before subtracting them (to guarantee that the difference will be strictly correct).

Commit e51b371 is the more dangerous one, because it replaces abs(unsigned1-unsigned2) with (unsigned1-unsigned2).  Per the discussion in #72, this is the wrong fix (even though Clang says to do it) because if unsigned1 is less than unsigned2 it will produce a nonsense result.

PR #118, opened 2019-05-08
Natmap

k5dat:
Removed 7 nonfunctional geo files for online maps from nationalmap.gov and added 5 new functional files.  Due to changes with respect to what is offered by nationalmap.gov, this is not an apples-apples update of the original 7 files. The new files contain "WMS" in the name because it is possible to create additional geo files that retrieve the same map data via tiles similar to OSM tiled maps.


godfreja:
Did you intend to close this without a merge happening?

tvrusso:
I will not be able to review this today.  Curt, could ya?  Looks really straightforward, but all 5 new maps need testing.

we7u:
Already checked out the maps earlier today. Work fine minus map timeouts since I have a large screen and the 'net is slow here. Cached maps are super-fast. These maps look great!

k5dat:
Thanks for merging it in. I had one timeout when testing them out this
morning. Wanted to be sure they still worked 'cause the files were created
back in November 2018.

Lee

PR #119, opened 2019-05-08
Issue117

we7u:
Removing PocketAPRS map format (*.pdb) from the Xastir code and docs.


we7u:
Looks like I branched off another issue branch so I got another commit in there I didn't want. Will resubmit in a bit.

Issue #117, opened 2019-05-08
Remove PocketAPRS map format (.pdb)


we7u:
I propose we remove the PocketAPRS map format entirely from the Xastir project (*.pdb files). It's doubtful anyone would still be using the format, but if so there are plenty of newer maps to choose from these days.

tvrusso:
Let's do it.  It is highly unlikely anyone is still using maps intended for turn-of-the-century PDAs.

Changing title.

we7u:
PR #120 deals with this.

PR #120, opened 2019-05-08
Remove PocketAPRS map format (*.pdb)

we7u:
Remove PocketAPRS map format from docs and code. This is an old map
format that is unsupported. No new maps are coming out for this
format and it's many years old now. If anyone is still using this
format they have newer and better alternatives.
Solves Issue #117


we7u:
Closes Issue #117

PR #133, opened 2019-05-23
Gcc8

we7u:
These commits solve a BUNCH of compiler warnings with gcc7 and gcc8, even with adding a bunch more compiler options like:

`CFLAGS="-O2 -Wall -Wextra -Wpointer-arith -Wwrite-strings -Wcast-align -Wbad-function-cast -Wformat-security -Winline -Wstringop-truncation -Wno-format-truncation -funsigned-char -W -Wno-discarded-qualifiers"`

These are perhaps the final commits to solving the compiler warnings in Issue #24

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,7 @@ matrix:
       dist: xenial
       name: "Ubuntu-16.04 gcc8 full-featured w/GraphicsMagick"
       env:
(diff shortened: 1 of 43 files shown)


we7u:
FYI: Compiling "master" right now with the same CFLAGS as listed above yields 879 warnings.

PR #134, opened 2019-05-24
Swap out dangerous string functions in callpass

we7u:
Flawfinder (Python) found dangerous string functions in callpass. The
code has been rewritten to avoid those functions plus formatted to our
new code formatting rules.


we7u:
Noted. Thanks.

Issue #136, opened 2019-06-04
lclint annotation: "/*@unused@*/" needed anymore?


we7u:
Years ago I added /*@unused@*/ in function declarations where the parameter wasn't used. This was for lclint checking of the code to get rid of hundreds or thousands of warnings. Do we need this in the code anymore? We have similar notations for the compiler now and those are more accurate. If people agree with removing them I'll yank them out of the code for good.

tvrusso:
I do not now use, nor have I ever used, lclint.  I'm cool with removing the
/* unused */ tags if you are.

we7u:
Closed by PR #137.

PR #137, opened 2019-06-04
Remove lclint annotation "@unused@"

we7u:
We don't currently use lclint, the compiler find unused parameters
anyway and we have notation already in the code for that. The lclint
annotations add extra clutter to the function declarations. These
annotations are out-of-date as well. Solves issue #136.


we7u:
It will. It already passed them on my branch. Will wait though 'cuz I think it shows failed builds if the PR gets merged before Travis CI builds are complete.

Issue #139, opened 2019-06-05
Remove unused directories Xastir/po and Xastir/intl  and all vestigial references to gettext and libintl


we7u:
I created a branch and deleted the "po" and "intl" directories from it. Xastir compiles/installs just fine. I've not ever touched this stuff, has anyone else? Can we get rid of them?

tvrusso:
These have been present in the repo since the very first "rc1" commit by Chuck Byam back in 2002.  They appear to be related to an old way of doing internationalization via GNU gettext and its libintl library.   Last references to the internal "intl" directory (in which intl/Makefile was configured) were removed from configure and various makefiles in commit c698d17.

Near as I can tell, this is dead code.

I am not sure what the "po" directory was for, but it appears also  to be related to internationalization with gettext. 

But in digging into this, I find that we still say we depend on libintl in configure.ac, but in fact our internationalization is not done using gettext at all, but using a simple system native to Xastir that has also existed since the very first "rc1" commit.  That code uses the functions in lang.c instead of anything provided by gettext.  Oddly, plenty of files in Xastir include "<libintl.h>" but do not appear to use anything provided by that header.

My guess is that not only do we not use the intl and po directories, we don't actually use gettext and libintl at all, even though configure probes for them *AND* we tell everyone we depend on it and we link against it.  I would bet we can remove the code that probes for (and adds to LIBS) libintl and libintl.h from configure.ac, and remove every instance of "#include <libintl.h>" from source code, and Xastir would work just fine with no impact on our internal internationalization stuff.

If you want to get rid of those two directories, you should probably also get rid of all the vestiges of Xastir's one-time use of gettext at the same time.

tvrusso:
Digging further:  The files that include libintl.h simply do this:

#ifdef  HAVE_LIBINTL_H
#include <libintl.h>
#define _(x)        gettext(x)
#else   // HAVE_LIBINTL_H
#define _(x)        (x)
#endif  // HAVE_LIBINTL_H


There is EXACTLY one place in which the "_(x)" macro is used, and it's in main.c:

t = _("X Amateur Station Tracking and Information Reporting");

This means that only on one place is there any use of gettext remaining in Xastir, and it's doing nothing because we no longer have any "po" files that actually handle translation of this string.

So if we remove all references to libintl.h from source, we also have to remove the "_()" macro usage from this one line of main.c.

we7u:
I see this also:
configure.ac:# Check for intl library (Needed for Cygwin)
[...]

PR #140, opened 2019-06-05
Po

we7u:
Closes issue #139 . Gets rid of Xastir/intl and Xastir/po directories, README-NLS file, references to libintl.h, etc.


Davis/Makefile.am
```diff
@@ -19,7 +19,6 @@ SUBDIRS = src
 #noinst_PROGRAMS = $(MyWX)
 
 
-# Remove this file here (it is created via configure), not from within intl.
 # DISTCLEANFILES =
```
godfreja: Should the DISTCLEANFILES line be removed as well?

we7u:
That's commented out anyway. I figured someone might want to use it again later and it doesn't hurt anything as-is.

Issue #116, opened 2019-05-08
Reformat sources to new coding standard


we7u:
Reformat Xastir sources per our new coding standards. The new standards were set May 7, 2019, 12:35 PM on the Xastir and Xastir-Dev lists after a vote:
```
2-space indents
Vertical brace alignment
Always use braces even when optional
```
See: http://xastir.org/index.php/HowTo:Developer_Guidelines_and_Notes
for more details. The Git project has Vim and Emacs config files added which help indentation.

NOTE: On hold until -AFTER- our next release.

we7u:
Not that this only applies to our *.c and *.h files, not our Makefiles or others.

we7u:
Looking at alert.c: Are we not going with the vertically-aligned braces for enums and similar too? I did for Davis/Lacrosse/callpass code.

tvrusso:
On Thu, Jun 06, 2019 at 09:55:27AM -0700, we recorded a bogon-computron collision of the <notifications@github.com> flavor, containing:
> Looking at alert.c: Are we not going with the vertically-aligned braces for enums and similar too? I did for Davis/Lacrosse/callpass code.

I didn't do that, but have nothing against it.  You can fix those if you like.
And I'll make sure to do it for the next round of files I attack.  I'm working
on awk.c and awk.h now, and I'll get those.

-- 
Tom Russo    KM5VY
Tijeras, NM  

 echo "prpv_a'rfg_cnf_har_cvcr" | sed -e 's/_/ /g' | tr [a-m][n-z] [n-z][a-m]

we7u:
See the response with 77 votes here:
https://stackoverflow.com/questions/2355834/how-can-i-autoformat-indent-c-code-in-vim
They use "astyle" from inside VIM. I may try that 'cuz it'll make a big job easier. At least if it can understand #ifdef's.

we7u:
So far it seems to work. Added this to Xastir/src/.vimrc:

" Using 'Artistic Style'or 'astyle' to autoformat/autoindent
" In VIM type 'gg' to go to the top of the file, then 'gqG'
autocmd BufNewFile,BufRead *.c,*.cpp,*.h set formatprg=astyle\ -A1\ -s2\ -S\ -xt0\ -xW\ -F\ -p\ -P\ -j

See http://astyle.sourceforge.net/astyle.html for details on the options.

we7u:
[...]

PR #145, opened 2019-06-10
Format code to new coding standards

we7u:
Per Issue #116 this changes to Allman style and 2-space indents, as
well as adding braces where optional.


we7u:
P.S. I'm holding off on other development (although I have a stack of O'Reilly X11 books I'm looking through at home) until we can get this or a similar pull request over with that reformats all of the code to the new coding standards.

Issue #148, opened 2019-07-16
Support ImageMagick 7


tvrusso:
ImageMagick 7 is not backward compatible to ImageMagick 6, even with the issue of #147 fixed.  The first, most obvious sign of this is that Xastir will not find "magick/api.h" at configure time, because they don't have that file anymore.  Xastir absolutely depends on it.

Supporting ImageMagick 7 will involve digging through Xastir's ImageMagick usage and figuring out whether we're using any features that have been completely removed, and if so, what we're supposed to do instead.  Some of this might have been taken care of by the older effort that fixed some of the deprecation warnings, but even if it was, we'll have to figure out the correct set of includes to use instead of the old "magick/api.h" include.

tvrusso:
Did a quick dig, and it looks like magick/api.h was also deprecated in ImageMagick6, and all it did was include magick/MagickCore.h.

It is entirely possible that this whole thing could be fixed by just probing for magick/MagickCore.h and including it if it exists, and only probing for magick/api.h and including it if the other fails.  There are a lot of places where this file is included, but if it is a correct fix, it's just a bunch of typing and a little more careful use of the macros that AC_CHECK_HEADER creates.

tvrusso:
Couple thoughts on this before I put it aside and go back to what I'm supposed to be doing today:

  - magick/api.h is tested for existence by calling AC_CHECK_HEADER, which only sets a shell variable use_imagemagick to yes or no depending on the result.
  - AC_CHECK_HEADERS can check for multiple headers and will not only execute the specified ACTION_IF_FOUND/ACTION_IF_NOT_FOUND, but also sets a macro "HAVE_<headername>" in config.h.

If we replace the AC_CHECK_HEADER call with an AC_CHECK_HEADERS that checks all of the appropriate files (magick/MagickCore.h and magick/api.h), we'll get a more complete picture of the Magick install and can better choose which thing to include, because we'll know which exists and which doesn't at the point of inclusion.  This is better than the simple yes/no "do we have ImageMagick" local variable (which we can still set and use in configure itself).

I hope to have some time to really play around with this later, but I had better knock it off for now.  I do believe that we have already gotten the code in proper shape to work with IM7 by getting rid of the IM6 deprecation warnings as part of issues #23 and #81.  I am hopeful that just fixing the include issue and the configure probe will be enough to close this.  But you never know with IM.

tvrusso:
I was overoptimistic about IM6 deprecation warnings in the previous comments.

[...]

PR #151, opened 2019-07-17
Remove ImageMagick version 6 deprecation warnings

### LINKED ISSUE #150: Remove all remaining ImageMagick 6 deprecation warnings
In the process of looking into issue #148, I realized that there was no hope of getting that bug fixed until we stopped using deprecated features of ImageMagick that were marked as such in ImageMagick 6.  All of those were REMOVED in IM7.

It is not enough to do this, because IM7 makes significant other changes that make it fail to build with Xastir.  But it is still good to remove the warnings.

tvrusso:
Even after the fixes of #23 and #40, we missed some ImageMagick deprecation warnings in main.c and map_geo.c.

This PR fixes those so that Xastir builds cleanly with ImageMagick 6.

I was unable to get ImageMagick 7 to build with these changes, and so put code into acinclude.m4 to reject versions of Magick with MagickLibVersion greater than 0x0700.  This may have been overly conservative, because the errors I saw when trying to build Xastir with my system's ImageMagick 7 might have been due entirely to the default package being built with HDRI support.

Anyone willing to start looking into bug #148 would be well-advised to make sure to start from a version if IM7 that has HDRI support disabled.

We should also probably put a test into configure that checks how IM was built and rejects it if HDRI support is on, rather than the blanket "don't use 7" I put in.  We can't work with ANY version of IM that has HDRI support enabled, not just 7.

Fixes #150.

### DIFF
--- a/acinclude.m4
+++ b/acinclude.m4
@@ -436,14 +436,37 @@ if test "${use_imagemagick}" = "yes"; then
   CXXFLAGS="$CXXFLAGS `${MAGIC_BIN} --cflags`" 
   LDFLAGS="$LDFLAGS `${MAGIC_BIN} --ldflags`" 
   LIBS="${MAGIC_LIB_DIR} `${MAGIC_BIN} --libs` $LIBS" 
-  # 
-  AC_CHECK_HEADER(magick/api.h, use_imagemagick="yes", use_imagemagick="no")
-  if test "${use_imagemagick}" = "no"; then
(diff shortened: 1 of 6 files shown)


tvrusso:
Since we're building Xastir with ImageMagick 6 in Travis, it should be sufficient review to see that these deprecations warnings are no longer thrown from main.c and map_geo.c.

PR #170, opened 2020-06-04
Have UDP server check that parent is still alive.

### LINKED ISSUE #130: Xastir can leave processes running if it crashes
If Xastir crashes in some way it can often leave a few processes running. These processes get in the way of starting up another Xastir later. Check whether it's possible to:
- Have Xastir look for such processes and kill them before fully starting up, or
- Have each process check periodically that the main process is still running, exit it not.

godfreja:
If xastir crashed it would leave child processes for the UDP server
still running. Switch the poll() to timeout after 5 seconds and then
use kill() to see if the parent is still alive.

Fixes #130.

### DIFF
--- a/src/x_spider.c
+++ b/src/x_spider.c
@@ -198,6 +198,7 @@ typedef struct _pipe_object
 } pipe_object;
 
 
+pid_t parent_pid;
 pipe_object *pipe_head = NULL;
 //int master_fd = -1; // Start with an invalid value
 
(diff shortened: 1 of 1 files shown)


tvrusso:
I'm willing to review, but am swamped at the moment and won't be able to look at it until at least the weekend.  I quickly looked over the code and it *looks* good, but I haven't tested it out.

godfreja:
No rush. I didn't think I was going to get to coding this until the weekend anyways.

godfreja:
Should I just go ahead and merge this or hold on for a review?

tvrusso:
I have not had time to review, and probably won't.

How have you tested this?  I tried just killing an Xastir process on my BSD system and it didn't leave a ghost process even before this fix, and I haven't had time to try to find another OS that would do so so I could check whether this branch fixed it.

I scanned the code and it looked reasonable, and I would not object to your merging it.  But maybe you should wait for Curt to weigh in.

godfreja:
I have ran it on my Mac, made sure I could still send data with xastir_udp_client, and verified that processes are not left around. (Without this mod processes were left around on MacOS,)
It would be nice to see this tried out on Linux, but I think the risk is low.

we7u:
I probably won't get to it until after Field Day, so feel free to
commit if you're happy with it.

On Sun, Jun 21, 2020 at 3:22 PM Jason Godfrey <notifications@github.com> wrote:
>
> I have ran it on my Mac, made sure I could still send data with xastir_udp_client, and verified that processes are not left around. (Without this mod processes were left around on MacOS,)
> It would be nice to see this tried out on Linux, but I think the risk is low.
>
> —
> You are receiving this because your review was requested.
> Reply to this email directly, view it on GitHub, or unsubscribe.



-- 
Curt, WE7U        http://xastir.org        http://www.sarguydigital.com

godfreja:
I created an Ubuntu VM and verified the fix behaves as I expect there.

PR #172, opened 2021-03-01
Use prefix path consistently for dbfawk test as well as download

Lexiyntax:
Fixes an issue where the dbfawk test fails on non /usr/local prefix paths by making the usage at the test step consistent with the download step based on the defined prefix path.


tvrusso:
Thank you for that.  That was quite an egregious error in the script to have ignored ${prefix} and used a hard-coded /usr/local/bin path.

Issue #174, opened 2021-03-05
relay_digipeat causes core dump


mobilinkd:
```
$ rpm -q xastir
xastir-2.1.4-8.fc33.x86_64
```

relay_digipeat causes Xastir to crash when compiled with stack fortification because:

```
sprintf(small_string,"relay_digipeat: outputs:\n\tport: %d\n\tcall: %s\n\tdest: %s\n\tpath: %s\n\tinfo: %s\n",
          port, call, destination, new_path, info);
```

Overflows the 200 byte buffer of small_string when trying to create the following string:

```
relay_digipeat: outputs:\n\tport: 6\n\tcall: K9IQP-13\n\tdest: APWW10\n\tpath: WX9O-3*,WIDE2-1\n\tinfo: }KD9QEN>APWW11,TCPIP,K9IQP-13*:}KD9QEN>APWW11,TCPIP,K9IQP-13*:@185408h4220.59N/08807.61W-PHG5100APRS-IS f
```

Which is 209 bytes long.

Backtrace follows.

```
#0  0x00007faa96c4d9d5 in raise () from /lib64/libc.so.6
#1  0x00007faa96c36954 in abort () from /lib64/libc.so.6
#2  0x00007faa9728090b in MagickPanicSignalHandler () from /lib64/libGraphicsMagick-Q16.so.3
#3  <signal handler called>
#4  0x00007faa96c4d9d5 in raise () from /lib64/libc.so.6
#5  0x00007faa96c368a4 in abort () from /lib64/libc.so.6
#6  0x00007faa96c90177 in __libc_message () from /lib64/libc.so.6
#7  0x00007faa96d214ca in __fortify_fail () from /lib64/libc.so.6
#8  0x00007faa96d1fd66 in __chk_fail () from /lib64/libc.so.6
#9  0x00007faa96c8893f in _IO_str_chk_overflow () from /lib64/libc.so.6
#10 0x00007faa96c94441 in _IO_default_xsputn () from /lib64/libc.so.6
#11 0x00007faa96c7e24b in __vfprintf_internal () from /lib64/libc.so.6
#12 0x00007faa96c889e4 in __vsprintf_internal () from /lib64/libc.so.6
#13 0x00007faa96d1f84f in __sprintf_chk () from /lib64/libc.so.6
#14 0x00005632636e7b7e in sprintf (
[...]

PR #175, opened 2021-03-05
Remove buggy, unused diagnostic code.

mobilinkd:
This code can cause a buffer overflow of small_string, which will
result in a core dump when Xastir is compiled with -fstack-protector.

https://github.com/Xastir/Xastir/issues/174

### DIFF
--- a/src/db.c
+++ b/src/db.c
@@ -21133,20 +21133,6 @@ void relay_digipeat(char *call, char *path, char *info, int port)
                       (unsigned char *)info,
                       strlen(info));
   }
-
-
-  //WE7U
-  sprintf(small_string,"relay_digipeat: outputs:\n\tport: %d\n\tcall: %s\n\tdest: %s\n\tpath: %s\n\tinfo: %s\n",
-          port, call, destination, new_path, info);
-  strncat(big_string,
-          small_string,
-          sizeof(big_string) - 1 - strlen(big_string));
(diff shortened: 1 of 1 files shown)


godfreja (approved):

tvrusso:
I think this change should go further, not just removing the last few lines of sprintfs into small_string and big_string, but the string variables themselves and all other sprintfs into them.  Just because these few lines were where you're seeing buffer overruns doesn't mean the other sprintfs aren't just as likely to be bad --- sprintf of a formatted string into a buffer with no length checks is a dangerous operation.

tvrusso (changes requested):
If we're going to remove any of these unused sprintfs and strncats, they should all go, along with the short char buffers that they write into.

mobilinkd:
While I agree with you, I am interested in a tactical fix for a crash that is affecting me.  I do not believe that a fix for this crash should be rejected because more work needs to be done.  I am happy to file another issue on that problem.  But don't expect me to make those changes.  I have no intention of doing so.

tvrusso:
> While I agree with you, I am interested in a tactical fix for a crash that is affecting me. I do not believe that a fix for this crash should be rejected because more work needs to be done. I am happy to file another issue on that problem. But don't expect me to make those changes. I have no intention of doing so.

Fair enough.

tvrusso (approved):

PR #187, opened 2021-11-06
Fix typing errors

dforsi:
Found using codespell.

### DIFF
--- a/Davis/src/db2APRS.c
+++ b/Davis/src/db2APRS.c
@@ -1371,8 +1371,8 @@ int main(int argc, char **argv)
     exit(10);
   }
   /* <dirkx@covalent.net> / April 2001 Minor change to allow quick
-   * (re)start of deamon or client while there are pending
-   * conncections during the quit. To avoid addresss/port in use
+   * (re)start of daemon or client while there are pending
+   * conncections during the quit. To avoid address/port in use
    * error.  */
   i = 1;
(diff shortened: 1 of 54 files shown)


dforsi:
Some changes in this PR are in printf strings, but most changes are in comments.
Let me know if you prefer to have two separate commits.
I tried to avoid changing third party code, like snprintf.c and I will amend this PR if you think that some changes should be reverted.

thanks,
Daniele

tvrusso:
I have reviewed these changes and most appear to be legitimate spelling errors.  The only exception I have found is in License where "du Pont" has been changed incorrectly to "du Point".

If you amend that I will review it again.

LICENSE
```diff
@@ -347,9 +347,9 @@ The licences which components of this software fall under are as follows.
 
 3)
   From 1991 to August 1999, ImageMagick was developed and distributed by
-  E. I. du Pont de Nemours and Company:
+  E. I. du Point de Nemours and Company:
 
-  Copyright 1999 E. I. du Pont de Nemours and Company
+  Copyright 1999 E. I. du Point de Nemours and Company
```
tvrusso, line 352: The company name is "du Pont" not "du Point" and so it is incorrect to "fix" this "error."

dforsi: Indeed, fixed now.

PR #194, opened 2022-05-15
configure.ac: egrep -> grep -E

thesamesam:
"egrep" is an obsolete alias for grep -E and newer greps will warn on usage
of egrep, so let's just swap it out.

Signed-off-by: Sam James <sam@gentoo.org>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -60,6 +60,8 @@ AC_PROG_LN_S
 AC_PROG_MAKE_SET
 AC_PROG_AWK
 AC_PROG_RANLIB
+# Used later on in our configure.ac for a nice summary
+AC_PROG_EGREP
 #AC_PROG_YACC
 
 AM_PROG_CC_C_O
@@ -885,7 +887,7 @@ AC_OUTPUT
 
 # Write a short summary of the configure tests/results to
 # "summary.log"
-egrep -e "checking |result:|gcc version" config.log >> summary.log
+$EGREP -e "checking |result:|gcc version" config.log >> summary.log
 
 
 echo ===========================================


tvrusso:
egrep is the traditional older alias, and Xastir is still used on old systems.

Let's not just swap it out.

thesamesam:
Because I was curious: it looks like `grep -E` was first required in POSIX.. 2001?

I have a better idea than just swapping it out anyway, let's use `AC_PROG_EGREP` which prefers `grep -E` but falls back to `egrep`. I'll update the patch.

tvrusso:
That is a much better solution.

I will wait for our macOS guy to review it before merging.

tvrusso:
I'm also asking @we7u to check it out to make sure it works on all of his systems, too.  I imagine it does, since Linux will definitely be posix compliant.

tvrusso:
While the current suggested commit does indeed remove the hard-coded egrep and replace it with a properly portable autoconfiscation mechanism, it seems the entire use of egrep here is not really useful.

The use of egrep here was inserted in commit fd3a1c61 simply to pare down config.log so that developers could more easily assist users when Xastir failed to configure for them.  That was in 2005.

In all the years I've been on the Xastir project, nobody has ever asked a user "send us your summary.log file" when trying to help diagnose a configure failure.  It's always been "send us your config.log file."  So clearly, nobody even remembers that configure is creating this summary, and it is somewhat pointless.  All the information is still there in config.log and those of us who do the diagnosing know how to read config.log.

So the *right* solution here may simply be to remove the ONLY line in all of configure.ac that uses egrep at all.

thesamesam:
I think that's a fair point. I try not to impose opinions like that when fixing bits like this in case it's there for a reason, but I _was_ wondering if it was really useful.

If none of you guys can remember it ever being useful, I'd dump it. I don't think I've seen this pattern in any other projects either.

tvrusso:
Interestingly enough, I just grepped configure for egrep after having removed the hard-coded egrep usage, and find that somehow AC_PROG_EGREP was already being called implicitly by some other macro that was present (I have not figured out which, because it's not us that's doing it, it's some other standard macro we're calling that pulls it in).  So $EGREP was already being defined correctly and could have been used.

But that's a little mystery for later.

thesamesam:
Thanks! Although good to not rely on implicit usage.

tvrusso:
Yes, I dislike depending on implicit calls like that and if we were to retain the egrep usage we should have inserted an explicit call like you did in this PR, but anyway, it's all moot now.

PR #200, opened 2022-11-06
Fix typing errors

dforsi:
Found using codespell:
codespell --skip *.dat,LICENSE,./src/snprintf.c,./scripts/langOldeEnglish.pl,./scripts/langPirateEnglish.pl --ignore-words-list 3st,bu,cancl,clen,herat,ists,lond,ro,templat,writen

### DIFF
--- a/Davis/README
+++ b/Davis/README
@@ -184,8 +184,8 @@ $ make
 $ su -c 'make install'
 
 
-ACKNOWLEGEMENTS & AUTHORS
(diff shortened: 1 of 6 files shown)


tvrusso:
Reviewed and merged.

Thank you for helping to keep us looking neat and clean here, and thank you for excluding some of the things that have been mistakenly flagged as "spelling errors" in the past.

PR #204, opened 2023-04-26
Fix C99 compatibility issue

djdelorie:
Related to:

  <https://fedoraproject.org/wiki/Changes/PortingToModernC>
  <https://fedoraproject.org/wiki/Toolchain/PortingToModernC>

Using return(0) instead of exit(0) means we don't need a prototype for exit() from stdlib.h.

Alternately, we could change sys/types.h to stdlib.h, and/or keep the return and remove sys/types.h.  Or just add stdlib.h.  It all depends on what, exactly, the original test was testing besides the obvious.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -647,7 +647,7 @@ LDFLAGS="-Wl,--no-keep-memory $LDFLAGS"
 AC_LINK_IFELSE(
  [AC_LANG_PROGRAM([[#include <sys/types.h>]],
                   [[/* Stupid useless test for linker flags */
-                     exit(0);]])],
+                     return(0);]])],
  [xa_cv_no_keep_memory=yes],
  [xa_cv_no_keep_memory=no])
 


tvrusso:
The entire purpose of that block of configure.ac was to check if the "-Wl,--no-keep-memory" loader option worked.  This was put in place in 2004 in commit 1de6009.  Use of "-Wl,--no-keep-memory" had been added unconditionally in commit  143bd5e as a quick hack to speed up Cygwin compilation just two weeks prior, without having first checked that it was actually usable.  And it turned out to have broken a few other systems whose compilers didn't support that option.

I believe using "return(0)" instead of "exit(0)" in this configure test won't break anything, and seems completely harmless to the level of backward compatibility this project tries to keep, while simultaneously addressing a forward compatibility issue.  The sole purpose of the test is to make sure that the test program compiles with the "-Wl,--no-keep-memory" option --- or rather, that the compiler doesn't *puke* on the "-Wl,--no-keep-memory" option.

But I'm asking that @godfreja and @we7u chime in before I just gleefully merge the pull request.

djdelorie:
Thanks!  As noted in the links, we're only really concerned with
implicit function declarations, as with gcc 13 those will cause hard
build errors.  Swapping exit() to return() is one of the standard fixes
we've been using (since return() doesn't need a prototype), but usually
those are ancillary to the thing being tested, not the test itself...

tvrusso:
I'm just going to go ahead and merge this, as it is a harmless change that can help others.

However, I think that at some point in the future we should remove this entire block of configure code, which serves to force "--no-keep-memory" into linker flags if the linker supports it.  This option tells the linker to optimize for memory usage instead of for speed, and should only be needed when the linker runs out of memory during the link phase.

Users should be expected to add options like this to their LDFLAGS when invoking configure if and only if they actually need it, configure shouldn't be forcing it in.  Especially when the options were put there primarily to support Cygwin 19 years ago.

The main problem I have with doing things like this in configure is that they are hard to maintain *and* can become obsolete very quickly.

Issue #203, opened 2023-04-16
update pcre support to pcre2


komacke:
It seems that the fedora maintainers can't update to xastir 2.1.8 due to an issue with pcre:

https://bugzilla.redhat.com/buglist.cgi?bug_status=NEW&bug_status=ASSIGNED&classification=Fedora&component=xastir&product=Fedora&product=Fedora%20EPEL

Could support be added so they can bump the package to the latest version?

tvrusso:
Development energy for Xastir is pretty low right now, and apparently the PCRE2 API is very much different from the PCRE we use in Xastir, so while it is certainly possible that we could do this, it is not a trivial request and might not be a real fast change.

There is no obvious "pcre to pcre2" migration documentation, and the original author of DBFAWK (the part of Xastir that uses PCRE) hasn't been around for a long time.  So whoever picks this up and works on it would first have to learn how DBFAWK uses pcre in detail, learn how to do the same thing in PCRE2's API, and do the coding to maintain backward compatibility (some systems that are NOT fedora continue to maintain PCRE 8.45 packages even though it's reached end of life already) with the aid of appropriate ifdefs and configure macros.

So:

  - I will take a look when I can
  - It would be phenomenally helpful if anyone else would like to pitch in
  - Don't hold your breath yet

tvrusso:
Since I'm looking at it now and scoping out the type of work that will be required, I'm going to ramble on what I've found.

The old PCRE API is documented here:
https://www.pcre.org/original/doc/html/

The new PCRE2 API is here:
https://www.pcre.org/current/doc/html/

The good news is that ALL pcre function calls in Xastir are in the file src/awk.c (with some data structures from pcre used in src/awk.h), and we only call the functions pcre_maketables, pcre_study, pcre_compile, pcre_exec, and pcre_free.

Flow of awk_compile_program in awk.c is this:
    - for each rule in an "awk program" linked list
       - clear out any existing parse table saved in the rule
       - allocate a table using pcre_maketables
       - use pcre_compile to compile that rule's pattern and store the resulting regexp in the rule's "re" element
       - optimize the regexp with pcre_study

The function awk_exec_program simply calls pcre_exec to check for matches with a compiled regexp.

There are functions similar to each of these in PCRE2, such as pcre2_maketables/pcre2_maketables_free, pcre2_compile/pcre2_compile_free, and pcre2_match.  The "optimization" of pcre_study is apparently now done by pcre2_jit_compile.

[...]

PR #206, opened 2023-07-26
Issue 203 -- Add support for PCRE2

tvrusso:
This pull request adds PCRE2 support to Xastir, while maintaining support for legacy PCRE.

If PCRE2 is found (first by finding "pcre2.h" and then by finding "libpcre2-8"), we use it.  Only if it is NOT found (or doesn't work because of pathing or other issues) then we probe for legacy PCRE as we always have.  If neither works, we puke saying we can't support shapelib without a version of PCRE and tell the user to install it or disable shapefile support.

I have tested this primarily by using the "testdbfawk" program over a handful of complex shapefiles I have.  I have confirmed that testdbfawk produces identical parsing and handling of shapefiles with either version of PCRE, and have tested the fallback mechanism multiple times --- Xastir compiles and links with either version of PCRE and works correctly.

I would like to merge this ASAP, as this issue (#203) is blocking some Linux distros from including Xastir in new releases (PCRE is deprecated and has been unsupported for years, whereas PCRE2 is still an active project).

Once merged, it will be a high priority to release a new version of Xastir, as the same distros that won't keep Xastir without PCRE2 support will not include anything but release versions.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -394,15 +394,31 @@ fi
 # Shapelib requires pcre now
 use_pcre=no
 if test "${shapelib_desired}" = "yes" -a "${use_shapelib}" = "yes"; then
-  AC_CHECK_HEADERS([pcre.h pcre/pcre.h],
-                   [AC_CHECK_LIB([pcre], [pcre_compile],
-                                 [use_pcre="yes(legacy)"
-                                  LIBS="$LIBS -lpcre"
-                                  AC_DEFINE(HAVE_LIBPCRE, 1, Define to 1 if you have the `pcre' library (-lpcre). )
-                                  AC_DEFINE(XASTIR_LEGACY_PCRE, 1, Define to 1 if you have the legacy `pcre' library (-lpcre). )
-                                  ])
-                    break
-                    ])
+  AC_CHECK_HEADERS([pcre2.h],
+                   [AC_CHECK_LIB([pcre2-8],[pcre2_compile_8],
+                                 [use_pcre="yes(PCRE2)"
+                                  LIBS="$LIBS -lpcre2-8"
+                                  ],
+                                 [use_pcre=no
+                                  AC_MSG_WARN([Found pcre2 headers but not libraries])
+                                 ])
+                   ],
+                   [use_pcre=no],
+                   [
+                     #define PCRE2_CODE_UNIT_WIDTH 8
+                   ])
+   if test "${use_pcre}" = "no"
+   then
+     AC_MSG_WARN([Modern PCRE2 not found, checking for legacy version])
+     AC_CHECK_HEADERS([pcre.h pcre/pcre.h],
+                      [AC_CHECK_LIB([pcre], [pcre_compile],
+                                    [use_pcre="yes(legacy)"
+                                     LIBS="$LIBS -lpcre"
+                                     AC_DEFINE(XASTIR_LEGACY_PCRE, 1, Define to 1 if you have the legacy `pcre' library (-lpcre). )
+                                                   ])
+                       break
+                       ])
+    fi
  if test "${use_pcre}" = "no"
  then
     AC_MSG_ERROR([Shapelib support requires PCRE, and we could not find PCRE.  Either disable shapelib by specifying "--without-shapelib" or install PCRE libraries and headers])
@@ -762,7 +778,12 @@ else
   ILIBS="$ILIBS         "
 fi
 if test "${use_pcre}" != "no"; then
-  ILIBS="$ILIBS PCRE"
+  if test "${use_pcre}" = "yes(legacy)"
+  then
+    ILIBS="$ILIBS PCRE"
+  else
+    ILIBS="$ILIBS PCRE2"
+  fi
 else
   ILIBS="$ILIBS     "
 fi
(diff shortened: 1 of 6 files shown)


configure.ac
```diff
@@ -394,14 +394,31 @@ fi
 # Shapelib requires pcre now
```
godfreja, line 394: The autoconf changes look OK to me, but I my experience of it is limited.

src/awk.h
```diff
@@ -22,11 +22,16 @@
...
-  #include <pcre/pcre.h>
+#ifdef XASTIR_LEGACY_PCRE
+  #ifdef HAVE_PCRE_H
+    #include <pcre.h>
+  #endif
+  #ifdef HAVE_PCRE_PCRE_H
+    #include <pcre/pcre.h>
+  #endif
+#else
+  #define PCRE2_CODE_UNIT_WIDTH 8
```
godfreja, line 33: Is this needed both here and in the autoconf rule?

src/awk.h
```diff
@@ -73,9 +78,15 @@ typedef struct awk_rule_
...
-  pcre *re;                   /* pcre compiled pattern */
-  pcre_extra *pe;             /* pcre optimized pattern */
+  #ifdef XASTIR_LEGACY_PCRE
+    const u_char *tables;       /* pcre NLS tables */
+    pcre *re;                   /* pcre compiled pattern */
+    pcre_extra *pe;             /* pcre optimized pattern */
+  #else
+    const uint8_t *tables;       /* pcre2 NLS tables */
+    pcre2_code *re;             /* pcre2 compiled pattern */
+    void *pe;                   /* pcre2 doesn't use separate optimization */
```
godfreja, line 88: I didn't see this being used in the PCRE2 cases, do we need to define it elsewhere?

godfreja (approved):
As a non-expert in both autoconf and pcre, it looks good. I will try this on my mac, but I may not get to it until tomorrow. It is up to you if you want to wait or not.

src/awk.h
tvrusso: Yes, it is needed in both places.

"pcre2.h" throws an error if you include it without defining PCRE2_CODE_UNIT_WIDTH.

The AC_CHECK_HEADERS macro tries to compile small test program that includes the header you request.  The macro is:
   AC_CHECK_HEADERS([header to check],[action if found],[action if not found],[extra stuff to define/include first])

The "#define PCRE2_CODE_UNIT_WIDTH 8" thing in that call *only* sticks that define in the little test program.

So if we need to include pcre2.h anywhere else, you MUST also define it before you include.

src/awk.h
tvrusso: Yes, we need to define "*pe" in the header even though it's never used, because there is an initializer in testdbfawk.h that tries to define a default set of rules, and it uses C initializer format --- so there's a value for pe in it (null).  If we don't have a place holder in the PCRE2 case, then we have to have two sets of initializers, one for PCRE2 and once for legacy PCRE.

If we remove PCRE support and go with ONLY PCRE2, we can remove "pe" from the struct *and* change the initializers so they don't try to initialize that element.

godfreja:
Thanks for the explanations.

I had a chance to try it out on my Mac, and was able to conpile/run using a simple shape file. I tried with both PCRE and PCRE2. (I did not try with neither.)

tvrusso:
The important thing to check is that Xastir builds and runs correctly, renders maps the way you expect them, and that you can use testdbfawk to probe how a given shapefile will render.

```
testdbfawk -f file.dbfawk -d file.dbf
``` 
will spew pages of data showing how file.dbf would be processed by file.dbfawk to standard err.  so if you compile and install Xastir using the PCRE build (e.g. if you don't have pcre2 installed or before checking out this PR branch) and then install PCRE2 and reconfigure/rebuild you should get the identical output.

Xastir itself will show "PCRE2" in the About box if it's using PCRE2 and "PCRE" if it's using the legacy version.   You will also get text output at the end of configure showing what PCRE is being used (either "pcre (Shapefile customization)...: yes(legacy)" or "yes(PCRE2)").

Another thing to look for is whether it fails to find pcre2.h even if you have it installed.  You *might* have to add a "-I" flag to CPPFLAGS to help us find it.  I don't think it should be necessary on most systems, where pcre.h and pcre2.h are both just dumped into /usr/include.  But some systems might put it into /usr/include/pcre2/pcre2.h or something.

The legacy pcre.h probe looks for it either at the top level of the current include search path or in a pcre subdirectory.  I did not emulate that because I hate that --- we shouldn't have to go searching all over the filesystem for what non-standard place it might have been installed, the search path should be augmented by the user when they have an oddball install.  This is done simply by adding `CPPFLAGS="-I/strange/path1 -I/strange/path2"` at configure time.

godfreja:
I'm guessing you typed up the testing comment before you saw my comment that I had tried it. For some details, I compiled with both PCRE and PCRE2 installed. I verified that PCRE2 was selected in the output of configure at the  end, then ran xastir and verified the simple shape file/dbfawk I use to show the bike route displayed correctly. I then uninstalled PCRE2 and recompiled. PCRE was used, everything displayed correctly. 
I did not need any additional configure flags in either case.

tvrusso:
Yes, I was writing that comment when I got an email saying there was a new comment on this PR.  I clicked "comment" before I read that email or refreshed the page.

df7cb:
Thanks from Debian for working on this! :+1:

Issue #25, opened 2018-02-16
Remove old ImageMagick API breakage workarounds (was "Remove ImageMagick support, require GraphicsMagick instead")


tvrusso:
Years ago, we floated the idea of dropping ImageMagick support in Xastir altogether, and just requiring GraphicsMagick.  ImageMagick's API used to change very frequently out from under us, and as a result the code is  peppered with ifdefs in order to test what version of ImageMagick is in use and to work around API instability.  These are all referencing the MagickLibVersion preprocessor macro.  Most reference versions prior to 0x0538-0x540.  It is highly unlikely that anyone even has those old versions anymore, and it may be that one would have to dig up a landfill to find a system that still ran such a version.  The current version I have installed (which is updated often) is 0x699.  The last time anyone added a backwards-compatibility ifdef was in 2003.

When that idea was floated, it turned out that there were a couple of Linux distributions that could not allow you to have both ImageMagick and GraphicsMagick installed from a package management system at the same time, because they both installed files of the same name into the same place.  I believe the last time we saw that was over 9 years ago.

Now, almost all distros and all OSen have sane packaging of these two products so they can coexist.  It may make sense now to remove the support for the highly unstable ImageMagick and just say we require GraphicsMagick full stop end of story.

Marking this as "Call for discussion" to give folks time to weigh in on whether they can reliably build Xastir with GraphicsMagick on their system, without interfering with other packages that require ImageMagick.  It may even be that nobody is using ImageMagick with Xastir anymore, because we've been telling folks not to for so long.

At the very least, even if we continue to support ImageMagick/GraphicsMagick interchangeability, it is very likely we can get rid of the 15-year-old ifdefs supporting versions prior to that date, simply for the sake of housekeeping.

dranch:
+1 for GraphicsMagick on Centos6 here

--David
KI6ZHD

we7u:
I much prefer GraphicsMagick over ImageMagick for a variety of reasons. GM is a fork of IM, but is more stable.

That being said, I find myself sometimes installing a Linux that doesn't have GM so I pick IM instead, knowing that Xastir will work with either. As for me, I'm willing to take on the
burden of installing from source for those cases where the Linux that doesn't have a GM package available. Most people would rather not install from sources though.

Will we be putting any burden on users if we force the use of GM and cut out IM support, for those that don't have GM as a standard package? That's about my only concern.

[...]

PR #209, opened 2023-07-27
Remove workarounds for ancient versions of ImageMagick

tvrusso:
In olden times, the ImageMagick API was quite unstable and kept changing out from underneath us.  Because it had not yet made it into package management systems, most users of Xastir that needed ImageMagick at the time *also* needed to compile it themselves from source.  As a result, many of them would not want to recompile the very latest version of ImageMagick every time an API got broken, but neither would new users want to have to stick to an old version. Xastir code was therefore peppered with all manner of ifdefs to handle every old version of ImageMagick one might have installed.

The first of these ifdefs was already present in 2002 when Xastir was first imported into CVS.  New ifdefs were added in 2002, 2003, and a final set in 2018.  There have been no more new versions of ImageMagick with broken APIs since then that we can use.

ImageMagick 6 is already ancient, and Xastir can't work at all with ImageMagick 7.

Maintaining ifdefs to support 20 year old versions of ImageMagick just bloats the code and makes it unreadable.  It isn't even necessary to continue to support those that were already old in 2018.

I have removed all the ifdefs that deal with any version of ImageMagick older than the one that reports its MagickLibVersion as 0x0669.  There have been no API breakages since then, and we can safely assume now that all users of ImageMagick 6 have a version at least as new as that.

ImageMagick support in Xastir is not long for this world, since ImageMagick's focus seems to be pushing harder into HDRI, which is completely incompatible with our image usage and would require a complete rewrite of the Magick code.  GraphicsMagick, however, continues to pursue stability of API rather than cutting edge image processing, and even if ImageMagick6 disappears from every single Linux package management system, we can still work so long as GraphicsMagick exists.

GraphicsMagick is already our recommended tool and we always use it if it is found and the user hasn't explicitly requested ImageMagick. We don't even look for ImageMagick unless the user has explicitly disabled graphicsmagick *or* graphicsmagick isn't found.

There are still a few ifdefs to deal with ONE GraphicsMagick API breakage, and someday we should probably remove that, too.  But issue #25 was strictly about removing "old ImageMagick API breakage
workarounds", so that's all I'm doing here.

Addresses #25

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -30018,15 +30018,7 @@ int main(int argc, char *argv[], char *envp[])
   InitializeMagick(*argv);
 #else   // HAVE_GRAPHICSMAGICK
 #ifdef HAVE_IMAGEMAGICK
-#if (MagickLibVersion < 0x0538)
-  MagickIncarnate(*argv);
-#else   // 0x0538 < MagickLibVersion < 0x0669
-  #if (MagickLibVersion < 0x0669)
-    InitializeMagick(*argv);
-  #else // > 0x669
-    MagickCoreGenesis(*argv, MagickTrue);
-  #endif
-#endif  // MagickLibVersion < 0x0538
+  MagickCoreGenesis(*argv, MagickTrue);
 #endif  // HAVE_IMAGEMAGICK
 #endif  //HAVE_GRAPHICSMAGICK
 
--- a/src/map_OSM.c
+++ b/src/map_OSM.c
@@ -543,28 +543,18 @@ static void draw_image(
   // try to reduce the number of colors in an image.
   // This may take some time, so it would be best to do ahead of
   // time if it is a static image.
-#if (MagickLibVersion < 0x0540)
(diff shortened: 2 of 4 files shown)


tvrusso:
Oh, and for what it's worth:  The last version of ImageMagick 6 is 6.9.12.  "0x0669" was in fact ancient, and it's not even possible to find a source code tarball that old on the official (ImageMagick 6 web site)[https://legacy.imagemagick.org/] anymore.  

So it is very safe to throw away all of this old code to support versions of ImageMagick that are ancient history.

godfreja (approved):
I don't have ImageMagick, so I can't do a test compile that actually tests these changes, but it looks good.

tvrusso:
I don't have IM 6 either, so I can't test it out, either.  However, all I did was remove ifdefs to support versions older than 6.6.9 (some as old as IM 5!), so anyone using any current ImageMagick would only be using the stuff I *didn't* remove anyway.  I'm gonna call it good.

PR #210, opened 2023-07-28
Issue 199 -- Move code to fix locate callback

godfreja:
The setup of the callback when an icon is clicked in the station list dialogs uses the callsign from the station name column to locate the desired station. The name was updated after the callback was set, so in a changing environment the callback would display the location of the station that previously occupied that row in the list.

Move the updating of the name to be before setting up the icon.

### DIFF
--- a/src/list_gui.c
+++ b/src/list_gui.c
@@ -757,6 +757,27 @@ void Station_List_fill(int type, int new_offset)
           ghost = 0;          // Not an object
         }
 
+        // call (or object/item name)
+        // Do this first as it is used for callback data later
+        /* check to see if string changed and over write */
+        temp_ptr = XmTextFieldGetString(SL_call[type][row]);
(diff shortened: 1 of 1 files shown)


tvrusso:
I will review this later this afternoon.

tvrusso:
Just tested this out.  Before fetching the updated branch I cleared all stations and opened the Last Stations dialog.  Let it populate a bit, then clicked one station.  The map centered on the *next* station in the list just as originally reported.

Repeated the operation with the fixed code and found that clicking on a station centered the map on *that* station, not some other.

tvrusso (approved):

tvrusso:
Please feel free to merge if you are ready.  I am not sure that Curt is available to do reviews.

tvrusso:
On second thought, I'm just going to go ahead and merge it myself right now.

PR #212, opened 2023-07-29
Documentation update

tvrusso:
In looking over the README.md, README.Getting-Started, and INSTALL documents before we release, I was dissatisfied with the layout and formatting.

This pull request reworks the INSTALL document into something that is in github markup (so it looks good when viewing in github), and a little more streamlined in the "how to install" department.

A lot of the text in the latter section, though, is directly copied out of README.Getting-Started, only formatted in markdown.  This is the stuff on running and configuring Xastir, and I think it was mostly fine as it was.

I have updated README.md to point to INSTALL.md instead of both README.Getting-Started and INSTALL.


tvrusso:
A couple comments:

The list of required and optional libraries in the original INSTALL document was bloated and included things that were merely dependent packages of other things we really require --- for example, Xastir does not itself use Freetype, something it pulls in needs that.

The reasoning there was that in the bad old days many users were building static libraries for our required packages from source code, and one had to include all manner of extra libraries to get things to link.  But now package management systems are almost universal in Linux and even Mac (macports or homebrew) and interdependency between packages is handled by that package manager.

So telling users which packages to install that we *actually* use is generally enough, the package manager will take care of the rest.

I have replaced specific reference to "XFree86" and "XFree86 development headers" and "XFree86 fonts" with "The X11 window system and its development headers", since nobody uses XFree86 anymore and "xorg" is the thing.  Still, even saying this may be too much, because any system with a desktop environment probably already has X installed.  But perhaps that ain't true for Mac.

tvrusso:
I recommend that to review this you look at the actual documents as rendered in github rather than as displayed as markdown source for the full effect.

Start here: https://github.com/tvrusso/Xastir/tree/documentation-update and look at the README.md it displays, then follow the link to INSTALL.md it provides.

godfreja:
I haven't had a chance to look at the updated docs yet, but thought I would answer comment about X on MacOS. Apple used to include an X server (XQuartz), but since MacOS 10.8 has not, and you have to download XQuartz yourself.

tvrusso:
Fair enough.  The INSTALL.md makes no specific claims about what is or isn't installed by default, and refers the user to system-specific guidance on the wiki (most of which is very outdated, but could be updated more easily by users than these documents).

The current wiki install notes for OS X under homebrew do tell the user to get XQuartz.

godfreja (approved):

Issue #198, opened 2022-09-23
Altitude-in-comments incorrectly delimited if station is transmitting PHG, does not get decoded as data and shows up just as a comment


westerkampj:
There appears to be a misplaced / in the interface.c code that cause the altitude to not be interpreted correctly and to appear at the beginning of the Comment in an APRS transmission. On line 7521, 

"A=%06ls/",

should be

"/A=%06ld",

I have made and tested this change and the altitude is interpreted correctly and displayed as an Altitude and no longer appears in the Comment.

Thanks!

John W.
W8LRJ

tvrusso:
Fascinating.  Yes, that *does* look like a typo, but  it's a bit puzzling that it matters due to the way that this is string is used:

```
      /* get CSE/SPD, Always needed for output even if 0 */
      xastir_snprintf(output_cs,
                      sizeof(output_cs),
                      "%03d/%03d/",
                      my_last_course,
                      my_last_speed);    // Speed in knots

      /* get altitude */
      if (my_last_altitude_time > 0)
        xastir_snprintf(output_alt,
                        sizeof(output_alt),
                        "A=%06ld/",
                        my_last_altitude);
...
```
and then later:
```
        if((strlen(output_cs) < 8) && (my_last_altitude_time > 0) &&
            (strlen(output_alt) > 0))
        {
          xastir_snprintf(output_brk,
                          sizeof(output_brk),
                          "/");
        }

        day_time = localtime(&sec);

        xastir_snprintf(data_txt_save,
                        sizeof(data_txt_save),
                        "@%02d%02d%02d/%s%s%s%s%s",
                        day_time->tm_mday,
                        day_time->tm_hour,
                        day_time->tm_min,
                        my_pos,
                        output_cs,
                        output_brk,
                        output_alt,
                        my_comment_tx);
```

This is *supposed to cause output_cs to end with a "/".  If the output_cs is not of the correct length, *and* the "output_alt" is fresh and non-zero length, then output_brk is supposed to contain "/".  In either case, this snippet is *supposed* to mean that there should be a slash before "A=" no matter what (either because it's already in output_cs or because we add one).

That is not to say I don't think your fix is correct.  It seems, though, to point to a different logic error that was clearly intended to take care of exactly this case.  Perhaps the conditional that sets output_brk to slash is slightly off?  So before I just go ahead and make this change, I'd like to hear from a few other users (paging doctors @we7u and @godfreja).

[...]

PR #213, opened 2023-08-04
Fix altitude-in-comment with PHG

tvrusso:
Xastir was improperly putting "A=xxxxxx/" into the comment field of position reports with PHG.  This is only done for fixed stations with GPS attached (the only ones that have altitude to report).  The reason it doesn't get done correctly is that there was a logic error in the code that was intended to insert a slash before the altitude in just the fixed-station+PHG cases.

It has been wrong since at least 2002.

The problem is that the spec requires "/A=xxxxxx" for the altitude to be recognized.

In other parts of the code, where we're adding altitude to course and speed, we were unnecessarily putting a "/" at the end of course and speed, then conditionally adding "/" in between that and the altitude string depending on whether it had already been added or not.  This is unnecessarily complicated, as the "/" should ALWAYS be attached to the altitude string, and in fact the spec does not even require that course and speed be followed by a slash.

The code that tried to insert the slash for PHG was broken and was not inserting one.

However, rather than fix the broken logic, I have chosen to simplify the code and keep it spec compliant.

This commit puts the slash at the beginning of the altitude string unconditionally, removes the unconditional slash from the end of course and speed, and removes all the attempted logic that was intended to insert the slash in between course/speed or PHG and the altitude.

I have confirmed that in every case where altitude was previously correct that it remains correct, and that in the case of PHG it is now correct where it had previously been wrong.

In doing so, I have also ripped out three identical copies of the code to format a fixed station report and refactored it into a function that is called three times.  Cut-and-paste code is bad form as it is harder to maintain three identical copies of code than one reusable element.

Addresses #198

### DIFF
--- a/src/interface.c
+++ b/src/interface.c
@@ -153,7 +153,14 @@ static int pop_count = 0;
 
 
 
+/* prototype for function that is only used in this file, need not be
+   in the associated header */
 
+void output_fixed_position(char *data_txt_save, size_t data_txt_save_size,
+                           char *my_pos, char * output_phg, char * output_alt,
+                           char *my_comment_tx,
+                           char *data_txt, size_t data_txt_size,
+                           char *output_net);
 
 // Fetch a record from the circular queue.
 // Returns 0 if no records available
@@ -7180,7 +7187,7 @@ void output_my_aprs_data(void)
   char output_phg[10];
   char output_cs[10];
   char output_alt[20];
-  char output_brk[3];
+
   int ok;
   int port;
   char my_comment_tx[MAX_COMMENT+1];
@@ -7451,7 +7458,6 @@ void output_my_aprs_data(void)
     output_cs[0] = '\0';
     output_phg[0] = '\0';
     output_alt[0] = '\0';
-    output_brk[0] = '\0';
 
 
     if (transmit_compressed_posit)
@@ -7484,7 +7490,7 @@ void output_my_aprs_data(void)
       /* get CSE/SPD, Always needed for output even if 0 */
       xastir_snprintf(output_cs,
                       sizeof(output_cs),
-                      "%03d/%03d/",
+                      "%03d/%03d",
                       my_last_course,
                       my_last_speed);    // Speed in knots
 
@@ -7492,7 +7498,7 @@ void output_my_aprs_data(void)
       if (my_last_altitude_time > 0)
         xastir_snprintf(output_alt,
                         sizeof(output_alt),
-                        "A=%06ld/",
+                        "/A=%06ld",
                         my_last_altitude);
     }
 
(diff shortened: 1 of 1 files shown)


tvrusso:
For the record, there are many special cases that needed to be tested before and after the hackage.  I did so by firing up Xastir and exercising each of those special cases to make sure that after it was all done the one special case that had been broken was no longer broken, and the other cases that had never been broken hadn't become broken.  here's what I found:

Before any hackage:
```
Case 4: fixed station, no weather
  no GPS, no weather, no timestamp, PHG disabled:
    =3458.19N/10618.30WlXASTIR-Linux
  no GPS, no weather, no timestamp, PHG enabled:
    =3458.19N/10618.30WlPHG2400XASTIR-Linux
  GPS, no weather, no timestamp, PHG enabled:
    =3458.19N/10618.30WlPHG2400A=007326/XASTIR-Linux
    THIS IS WRONG

Case 1:  Mobile, with local time (PHG is never transmitted):
  no GPS:
   @040943/3458.20N/10618.30Wl000/000/XASTIR-Linux
  GPS:
   @040943/3458.19N/10618.30Wl327/000/A=007347/XASTIR-Linux

Case 2:  APRS mobile with Zulu time:
  No gps:
   @041545z3458.19N/10618.30Wl000/000/XASTIR-Linux
  GPS:
   @041546z3458.19N/10618.31Wl304/000/A=007360/XASTIR-Linux

case 3: APRS mobile, zulu, seconds
  No GPS:
   @154727h3458.19N/10618.30Wl000/000/XASTIR-Linux
  GPS:
   @154802h3458.19N/10618.30Wl088/000/A=007344/XASTIR-Linux
```

After refactor of cut/pasted code into reusable function:
```
Case 4: fixed station, no weather
  no GPS, no weather, no timestamp, PHG disabled:
   =3458.20N/10618.29WlXASTIR-Linux
  no GPS, no weather, no timestamp, PHG enabled:
   =3458.20N/10618.29WlPHG2400XASTIR-Linux
  GPS, no weather, no timestamp, PHG enabled:
   =3458.19N/10618.30WlPHG2400A=007337/XASTIR-Linux


Case 1, 2, 3 unchanged code, didn't bother recapturing.
```

After removal of unnecessary output_brk complexity:
```
Case 4: fixed station, no weather
  no GPS, no weather, no timestamp, PHG disabled:
   =3458.19N/10618.30WlXASTIR-Linux
  no GPS, no weather, no timestamp, PHG enabled:
   =3458.19N/10618.30WlPHG2400XASTIR-Linux
  GPS, no weather, no timestamp, PHG enabled:
   =3458.19N/10618.30WlPHG2400/A=007355XASTIR-Linux
    THIS IS CORRECT

Case 1:  Mobile, with local time (PHG is never transmitted):
  no GPS:
   @041017/3458.19N/10618.30Wl000/000XASTIR-Linux
  GPS:
   @041018/3458.20N/10618.30Wl186/000/A=007254XASTIR-Linux

Case 2:  APRS mobile with Zulu time:
  No gps:
   @041619z3458.19N/10618.30Wl000/000XASTIR-Linux
  GPS:
   @041623z3458.20N/10618.30Wl223/000/A=007285XASTIR-Linux

case 3: APRS mobile, zulu, seconds
  No GPS:
   @162431h3458.19N/10618.30Wl000/000XASTIR-Linux
  GPS:
   @162504h3458.19N/10618.31Wl090/000/A=007329XASTIR-Linux
```

godfreja (approved):

PR #215, opened 2024-02-17
Update to latest version of the NWS shapefiles

nq0m:
Updates the get-NWSdata script to the current versions of the shapefiles


nq0m:
I think this one is pretty straightforward for you guys.  I noted that the filenames in the get-NWSdata script were out of date, so I went and found the current versions of everything and created this PR.

I also note that it looks like every one of these files also has an upcoming new version on March 5.

tvrusso:
Looks good.  Tested and verified, merging.

tvrusso:
Thank you.

Issue #217, opened 2024-05-27
Update WMS map caching capability to use more recent version of Berkeley DB


tvrusso:
The code in Xastir that uses Berkeley DB for map caching is rather tightly tied to Berkeley DB (BDB) version 5 (DB5), which is ancient, deprecated, and beginning to disappear from package management systems for lack of support.

This mpa caching code was originally put in place to support TIGER online maps, which went away years ago (and support for which was removed from Xastir in 2019 anyway).  But the caching of map images from WMS servers (accessed via .geo files) is still supported and disabling it would make use of WMS servers for mapping very much slower and more wasteful.

Most recently, the maintainer of the FreeBSD port of Xastir has proposed removing the use of Berkeley DB from his port (simply by disabling it being probed and built against) because FreeBSD wants to drop Berkeley DB5 from their package system.

It would probably be great if we could make Xastir work with more recent versions of BDB.  DB 18 is the current version.

Even if the API has not changed at all, the method that Xastir's configure script uses to detect DB and add CFLAGS and LIBS is very kludgy and very much expecting only DB5 to work.  That code was swiped from another project many years ago and explicitly loops through specific versions of DB to look for, going from the more recent to the much older all the way back to DB 4.0.  In fact, it is often necessary to bypass that code and tell Xastir exactly where to look for headers and libraries  using the `---with-bdb-incdir` and `--with-bdb-libdir` options anyway, because the kludged code is coded to look in specific places appropriate to only one or two operating systems.  Further, it only checks to see if it can link to libraries with specific version numbers in their names, so even if one points it at the db18 *headers* it will fail because it won't be looking for the correct corresponding library (and will find db5's library if it exists, then complain about version mismatch).

The configure code that would need to be fixed is in acinclude.m4, the XASTIR_BERKELEY_DB_CHK macro and a few other macros it calls (all with BERKELEY in their names).   The only file that actually contains Berkeley DB calls is in map_cache.c, which is the only one that includes the BDB-related `db.h` include file.

It is entirely possible that the only thing that needs to be done here is to fix the probing for libraries and headers so that more recent versions can be used, but one would have to study whether and how the BDB API has changed since version 5 before concluding that.

tvrusso:
FWIW, it looks like the only Berkeley DB functions we use are db_create (to get a  pointer toi data structure for controlling a database, dbp), and then dbp->open(), dbp->get(), dbp->put(), dbp->close(), dbp->err(), and db_strerror.

[...]

PR #218, opened 2024-05-27
Add detection and use of db18

tvrusso:
As a very, very early attempt at addressing Xastir/Xastir#217, I have tweaked the very, very simplistic autoconf check for DB libraries to look for libdb-18.1 or libdb-18 in addition to all the libdb-5 checks we already do, placing the newer version earlier in the detection loop.

I have confirmed that libdb18 has changed absolutely nothing in the API that breaks Xastir's use, and in fact by simplyt telling my build to look for db-18 headers in /usr/local/include/db18 and libraries in /usr/local/lib, the DB18 libraries are successfully linked in and work correctly.

However this is not ideal, as if one tells Xastir to look for db5 headers and both db-18 and db-5 libraries are installed in /usr/local/lib, then Xastir uses the headers from db5 and the library from db18, with consequent failure at run time (with a warning and disabling of DB use:

***** WARNING *****
Berkeley DB header files/shared library file do NOT match! Disabling use of the map cache.
 Header file: Berkeley DB 5.3.28: (September  9, 2013)
Library file: Berkeley DB 18.1.40: (May 29, 2020)
***** WARNING *****

So the configure detection really ought to be beefed up to make sure that headers and libraries match just as the code does.  But because we only *compile* a configure check and don't run it (which is the right way to do it lest one break cross compilation), we *can't* do that check, which involves calling db_version and checking its result.

In fact, the entire reason we *have* that warning is that users have this problem all the time and this is the only way to tell them that they must fix their configure arguments.

It is entirely possible that this is the best we can do at the moment, just as it was the best we could do for other BDB version issues.

I have ONLY added checks for libdb-18.1 and libdb-18, I have not tried to add other versions in between version 5 and 18.

This at least partially addresses Xastir/Xastir#217

### DIFF
--- a/acinclude.m4
+++ b/acinclude.m4
@@ -999,7 +999,7 @@ AC_DEFUN([XASTIR_BERKELEY_DB_CHK_LIB],
 # to specify the function call with the full prototype for it to be found.
         BDB_LIB_FOUND="none"
         AC_MSG_CHECKING([for a library containing db_create])
-        for dbname in db-5.3 db5.3 db53 db-5.2 db5.2 db52 db-5.1 db5.1 db51 db-5.0 db5.0 db50 db-4.9 db4.9 db49 db-4.8 db4.8 db48 db-4.7 db4.7 db47 db-4.6 db4.6 db46 db-4.5 db4.5 db45 db-4.4 db4.4 db44 db-4.3 db4.3 db43 db-4.2 db4.2 db42 db-4.1 db4.1 db41 db-4.0 db4.0 db-4 db40 db4 db
+        for dbname in db-18.1 db-18 db-5.3 db5.3 db53 db-5.2 db5.2 db52 db-5.1 db5.1 db51 db-5.0 db5.0 db50 db-4.9 db4.9 db49 db-4.8 db4.8 db48 db-4.7 db4.7 db47 db-4.6 db4.6 db46 db-4.5 db4.5 db45 db-4.4 db4.4 db44 db-4.3 db4.3 db43 db-4.2 db4.2 db42 db-4.1 db4.1 db41 db-4.0 db4.0 db-4 db40 db4 db
           do
 	    LIBS="$saved_LIBS -l$dbname"
       AC_LINK_IFELSE(


tvrusso:
The "runtime detection of mismatch" thing is a very old issue in Xastir and used to bite users all the time back when both db5 and db4 tended to be in package management systems.   

While it is undesirable, it is at least not a new annoyance.  I would not consider it to be a blocker to accepting this PR, but perhaps other developers might hate it.  I would love to see our BDB detection be less primitive, but as it stands I can't be the one to improve it right now.

tvrusso:
This PR might require additional hacks for other operating systems.  The crude loop over versions looks for libraries with specific names, and in older versions we looked for things like db-5.3, db5.3, and db53 to cover all the bases.

In fact, the entire loop should be preceded by a test to see if we find db_create *without* any addition to LIBS, to allow users to *force* configure to use a specific named library using "LIBS=-l" flags, and not to bother doing the loop if the link succeeds without any automated LIBS addition.

tvrusso:
My additional commit to this branch adds the "check if db_create is already in LIBS" check mentioned above.

I have tested that if I don't specify LIBS then the latest db library version is always used (irrespective of whether it matches the headers provided), but if I specify a `-ldb-5.3` in LIBS on the configure line it is used without the loop ever being entered.

godfreja (approved):
I tried this out on my Mac and it seemed to work. I had to point configure at the include and library directories for Berkley db 18 using CPPFLAGS and LDFLAGS. This is not new, I had to do this previously with version 5 as well due to the way homebrew handles Berkley DB installs. 

I'm only using OSM and a handful of shape files for maps, so I didn't exercise the caching code directly.

As far as the code changes, they look OK to me, but I have limited experience with autoconf.

tvrusso:
An easy test is to use the  WMS_USGS_Topo.geo online map and watch to see that it's caching maps as they are loaded, then reusing them from the cache as you pan around and back to places you've already been.   There is a display in the status line about whether it's using a cached map or if a map is not found in cache.

That's the one I used to check this out with.

godfreja:
I am seeing appropriate caching behavior with WMS_USGS_Topo.geo.

tvrusso:
I'll give it a day to see if @we7u wants to check it out, and will merge if he approves or doesn't comment.

tvrusso:
But since I see no position reports from Curt in quite some time I am guessing he's out of contact these days.

tvrusso:
I should note that many of the changes in the diffs here are whitespace changes because my editor was failing to recognize matching braces in the presence of the include directives inside lines.  Adding the whitespace and putting the includes on separate lines fixed that issue and made it easier to spot correctly nested braces.

Most of the meat is the addition of "db-18.1" and "db-18" to the for loop and addition of a block of `AC_CHECK_LIB`, and wrapping what had been there before in a conditional so it is only probed if the earlier check doesn't find db_create.

tvrusso:
One last comment from me before I stop.

I just tried building Xastir on my Ubuntu 22.04 machine only to find that at least that version of Ubuntu *only* has db5.3 (5.3.28) with no more recent option available.  Near as I can tell even the later version of Ubuntu doesn't have db18, so I can't test whether this hack works there with the new version.

But it does correctly use db 5.3 in exactly the way it used to, so at least I didn't break anything.

It doesn't look like anybody's spending any time getting more recent Berkeley DB working on that system.  It's possible they're considering its license incompatible with Linux.  Shrug.

PR #219, opened 2024-05-28
Greatly simplify Berkeley DB library search

tvrusso:
Way back in 2004 when map caching was added to Xastir, a set of autoconf macros was copied from another project to detect which Berkeley DB libraries to use.  This set of macros was kept for the last 20 years, with only minor tweaks.

However, the thing it did was very much similar to the standard "AC_SEARCH_LIBS" autoconf macro, but clumsier.

It was incorrectly assumed that one must include "db.h" in the test program for db_create, which means one must also provide a test program that uses db_create with all its required arguments --- something AC_SEARCH_LIBS cannot do.

This commit vastly simplifies the XASTIR_BERKELEY_DB_CHK_LIB macro to rip out all the convoluted looping and conditionals, and replace it all with a call to AC_SEARCH_LIBS with a big list of possible libraries.

AC_SEARCH_LIBS does everything that the old loop had done in a more readable fashion.  Further, the very first thing that it does is to try to link a test program containing the db_create function *without* adding any libraries to LIBS --- thus doing exactly what my earlier commit did: allow the user to specify the exact library to use on the configure line (e.g. by using LIBS="-ldb-5.3") and skip the loop.

I have removed all of the DB4 libraries from the search list.  DB5 is itself very old, and DB4 is no longer in any package management systems.  The license changed with DB6 and a lot of Linux systems don't even have that, and Oracle is already up to DB18.  FreeBSD has DB18 in its ports tree, and has deprecated DB5 (the deprecation of DB5 and removal of Xastir's dependency on DB5 is what prompted all of this activity on DB detection).  There is no point in having Xastir's configure even bother looking past db 5.0 anymore.

Xastir/Xastir#217

### DIFF
--- a/acinclude.m4
+++ b/acinclude.m4
@@ -926,60 +926,34 @@ fi
 ])
 ])
 
-# From Cyrus imap distribution (KB3EGH)
-
-dnl These are the Cyrus Berkeley DB macros.  In an ideal world these would be
-dnl identical to the above.
-
-dnl They are here so that they can be shared between Cyrus IMAPd
-dnl and Cyrus SASL with relative ease.
-
-dnl The big difference between this and the ones above is that we don't assume
-dnl that we know the name of the library, and we try a lot of permutations
-dnl instead.  We also assume that DB4 is acceptable.
-
-dnl When we're done, there will be a BDB_LIBADD and a BDB_INCADD which should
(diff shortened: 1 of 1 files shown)


godfreja (approved):
I like seeing all that red in a PR diff.

Tested this out on my Mac, worked fine.

PR #221, opened 2024-07-24
Add a better quality PNG icon for Xastir

### LINKED ISSUE #220: Add a better quality desktop icon
The [current desktop icon](https://github.com/Xastir/Xastir/blob/master/symbols/icon.png) is too small for today's standards/package managers and causes Xastir to be hidden from software centers in Ubuntu and other Linux distributions. Please, consider adding at least a 128 x 128 px png file or a scalable svg one.

This is a svg file that we made for our Fedora Linux Xastir package:

xastir

AsciiWolf:
This adds a standard 128 x 128 px icon upscaled from the original one.

Closes #220.

/cc @tvrusso


tvrusso:
I have a counter proposal for you.  Rather than replacing the very old, 40x40 icon completely, move it to icon_small.png, add your new, larger icon as icon.png, and add your icon.svg as well.  That way users of ancient systems can keep the old, tiny one (we still cater to those users, as hams love to take computers out of the garbage and repurpose them for hammy things), users of modern systems and package managers can use the big one, and the SVG will be available for anyone who wants it.

tvrusso:
On second thought after some testing I find that having the larger icon changes nothing about my existing desktop (I was concerned it would be too large), as the desktop manager scales the icon to a uniform size anyway.  So my concerns about catering to older systems are unfounded, and I just went ahead and merged this PR.

Thank you.

If you want to open a new PR to add an SVG version, I'd gladly except it, too.

PR #224, opened 2024-08-02
Delete get-pop and update get-NWSdata

### LINKED ISSUE #35: get-pop
The get-pop script is returning a 404.

godfreja:
Remove the get-pop script as it points to data that is no longer there.

Update the get-NWSdata script to point at the current versions of the files. New versions go into effect on September 10th and the script should be updated again then.

Closes #35.


tvrusso (approved):
These look like the right things to do in every regard.

get-pop is removed as obsolete and broken, get-NWSdata is updated to reflect new file names.  I have verified that the new shapefiles do not require new dbfawk files, as the files have not changed DBF signature.

PR #232, opened 2025-10-02
Fix (possible) spaces in WMTS-URI

DJ3CE:
The formatting of the bounding box (`BBOX`) left spaced in the resulting URI, due to formatting with `%8.5f`, e.g., where a `5.2` will result in ` 5.20000` (mind the space in first place to gain 8 characters) and leading to cURL-errors due to malformed URIs.

This commit fixes the issue by using leading zeros (`%8.5f` --> `%08.5f`) in the format specifications.

### DIFF
--- a/src/map_WMS.c
+++ b/src/map_WMS.c
@@ -356,7 +356,7 @@ void draw_WMS_map (Widget w,
 
 // This specifies a bounding box based on square pixels.
   xastir_snprintf(tmpstr, sizeof(tmpstr),
-                  "&BBOX=%8.5f,%7.5f,%8.5f,%7.5f",
+                  "&BBOX=%08.5f,%07.5f,%08.5f,%07.5f",
                   left,   // Lower left
                   bottom, // Lower left
                   right,  // Upper right


tvrusso:
I agree, it looks like the code *should* misbehave on any lat/lon box that has only one digit of latitude or two digits of longitude, and making it output with zero padding instead of spaces as you have done should be the correct fix.

This certainly looks like a correct fix to me.   Odd, though, that it hasn't been noticed in the 20 years since that file was created and that line added, so I'd like to see an actual instance where it breaks so we can test the program before and after this fix.

Do you have a WMS .geo file and bounding box where the existing code actually misbehaves?   Could you add it to this discussion so we can do proper testing?

tvrusso (approved):
Tested by using Online/nationalmap.gov/WMS_USGS_Topo.geo and an appropriate window with problematic bounding box (0.60698,-17.31992,46.58273,1.51195).   This wound up with a space before one of the elements of the bbox, and the proposed changes fixed it.

PR #233, opened 2025-10-02
Add GitHub Actions to test build for linux MacOS

godfreja:
Use GitHub actions to do a test build on ubuntu (both a smaller scale and full-featured version, minus kernel AX25) and on macOS.

### DIFF
--- a/.github/workflows/linux-build.yml
+++ b/.github/workflows/linux-build.yml
@@ -1,4 +1,4 @@
-name: Basic Ubuntu Build
+name: Test Builds
 
 on:
   push:
@@ -7,8 +7,7 @@ on:
     branches: [ "master" ]
 
 jobs:
-  build:
-
+  basic-ubuntu-build:
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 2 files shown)


tvrusso:
Only one comment, since I don't know github actions well and can't verify or help much:

"bootstrap.sh" is sorta doing exactly the same thing that "autoreconf -i" would do, but with more noise and less clarity.  You might want to change all of those runs of bootstrap.sh to `autoreconf -i`, or maybe one of them, to keep sanity checking the more modern approach.

One package system I know of had to use autoreconf instead of bootstrap because it fits better with their automated packaging system.  That approach was very temporarily broken years ago by someone inserting non-autoconfy things into bootstrap.sh that should have been (and very quickly *were*) done instead in a Makefile.  Having automated testing check the autoreconf approach would help identify just such an issue if it were to recur.

The autoreconf approach is currently mentioned only in README.developers.md, and probably should be more prominently documented.

tvrusso:
You might also consider doing a squash-and-merge rather than preserving all 9 commits as separate.  But that's just a housekeeping call for you to make.

Issue #239, opened 2025-10-31
Selecting an Overlay Char in range '0' to '9' and Compressed Packet results in faulty packet


F4FXL:
If the user selects an overlay char between '0' and '9' and enables packet compression, the resulting packet is faulty and is not decodable by any other APRS software/device.
Selecting the regular \ symbol, the packet is valid.

<img width="699" height="224" alt="Image" src="https://github.com/user-attachments/assets/257254c7-c65b-4fc2-9633-3350225a492b" />

tvrusso:
Is even xastir itself unable to display this properly?

This might take a little digging.

Thank you for the report, we'll have to look into this.

tvrusso:
Deleting that object was problematic through the GUI, since the received posit couldn't be displayed, centered on, or selected.  I had to delete it from object.log and restart Xastir.

The uncompressed object posit was fine:

`;Frobnitz *311952z3459.04N310611.07W-Testing uncompressed posit`

tvrusso:
Just enabling some debugging output, the group character is corrupted by the time the `compress_posit` function is called.  Going to have to work backwards to figure out where it's going wrong.

tvrusso:
The frustrating bit is that the code for *calling* compress_posit is scattered all over the place and there are multiple paths the code takes when creating an object (the OP is using compressed posits on the station itself, but the bug exists for objects, too).   Much of that code is cut/paste duplicated across multiple functions and conditionals, and the code that *creates* an object is different than the code that maintains it.   And it appears that at least the code that creates the initial posit is already producing a bad packet, which gets inserted into the database by the code that *processes* incoming packets, which then corrupts it further.   Worse, some of the paths interact directly with gui dialogs, and it'll be difficult to disentangle them.  I fear that trying to create unit tests will be difficult.

And that's different from the code that does the main station posits, which is more duplicated code with apparently duplicated bugs.

It's going to take a while to wrap my brain around what the code is doing and I'm backed up with projects at the moment.  I'll have to block off time to do some serious debugging and tracing of all the different paths to figure out where it's initially getting off the rails.   I might not get to it until well into next week.  

If you care to look into it, great, but it seems to me that your work on the DB and unit testing thereof seems like enough work for you already.

tvrusso:
> Thanks for the replies. I am in no way familiar with the internal code structure of Xastir.... I might try my chance, but will take a while.

[...]

PR #240, opened 2025-11-01
Overlay chars 0 to 9 are encoded a to j #239

F4FXL:
This is a quick fix for #239

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -2157,7 +2157,7 @@ char *compress_posit(const char *input_lat, const char group, const char *input_
   xastir_snprintf(pos,
                   sizeof(pos),
                   "%c%s%s%c%c%c%c",
-                  group,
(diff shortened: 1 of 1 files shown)


tvrusso:
I would just like to comment that this pull request *would* be the right thing to do if there wasn't so much other code that is doing the wrong thing.

It is absolutely correct to transmit a-j instead of 0-9 for numeric overlays in compressed mode, and that is exactly to spec as noted in #239.   

Xastir *should* be doing the 0-9 -> a-j conversion in the compress_posit function as is attempted here, but there are tons of places in objects.c that are trying to do it already (and doing it wrong) so that if we did what was suggested here *now* other things would break badly (worse than they're already broken).

The problem that you're fixing this way only fixes station posits, which aren't mangling the overlay character already.

Issue #257, opened 2025-11-13
Split up "util.c"


tvrusso:
Util.c has become something of a catch-all file for random functions that seemed to have no other home.

It is *mostly* functions that stand alone and work only on their arguments, but some functions make use of global variables from main.c or elsewhere, meaning that the file has become intimately intertwined with other files, making it *really* difficult to unit test anything in it or anything that makes use of anything in it.

I propose we split util.c into multiple files to reduce this interdependency.  

util.c should contain all of the functions that are truly stand-alone and self-contained, depending on no global data.   

Things like hash table functions (which make use of global variables defined in util.c itself, but are called only in db.c and main.c) and other functions that aren't self-contained should be split off and live in other .c and .h files where they are appropriately grouped.

Doing this will reduce the bloat involved in including "util.h" in any other file, and make util.c really just be a file of basic utility functions.

PR #263, opened 2025-11-23
Add a test suite for the util.c functions

tvrusso:
In my work on #257 I tried to pick util.c apart to the point that it no longer was a huge pain to make use of some of its functions in the unit test suite.

This PR adds a tiny set of unit tests of the simplest functions in util.c, with all non-essential external functions stubbed out to satisfy the linker.

Since I need to use the functions I'm testing here in some of my other refactors, I want to get them under unit testing before I proceed.

Related to #257


tvrusso:
This PR is so straightforward it is not worth wasting anyone else's time requesting a review.  My main reason for submitting it was to put it through the CI actions and make sure it all works correctly on every platform.

I'm merging it myself.

PR #267, opened 2025-11-27
Refactor object/item deletion

### LINKED ISSUE #265: Refactor object/item deletion code
The functions `Object_change_data_del` and `Item_change_data_del` call `Setup_object_data` and `Setup_item_data` to get a packet for a normal object or item, then change one character of the string returned to '_' to signify that the packet is killed.

However, in my work on #249 and #263, I now have the ability to set a packet as "killed" directly on creation.

It should be a simple matter to change `Setup_object_data` and `Setup_item_data` to take an extra argument ("killed") to pass down to `construct_object_item_data_row` and get the packet correct without extra, repetitive code.

tvrusso:
This PR refactors the coding of two object/item deletion functions so that the actual code that marks a packet deleted is in exactly one place (`Create_object_item_tx_string`, where it already was coded and heavily unit tested).

This PR is a WIP because I will be rebasing this branch off of master after #266 is merged.  As it stands now it appears to include all the commits of that other PR.

Closes #265

### DIFF
--- a/src/objects_gui.c
+++ b/src/objects_gui.c
@@ -134,7 +134,7 @@ int Read_object_item_dialog_values(char *name, size_t name_size,
                                    char *prob_max, size_t prob_max_size
                                    );
 int Setup_object_item_data(char *line, int line_length, DataRow *p_station,
-                           int is_object);
+                           int is_object, int is_killed);
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 
 // Init values for Objects dialog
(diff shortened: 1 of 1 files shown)


godfreja:
I'm not seeing is_killed being used.

src/objects_gui.c
```diff
@@ -385,7 +385,7 @@ int Setup_object_item_data(char *line, int line_length, DataRow *p_station,
                                               bearing, NRQ,
                                               prob_circles,
                                               prob_min, prob_max,
-                                              is_object, 0);
+                                              is_object, is_killed);
```
tvrusso: is_killed is getting used right here.

godfreja:
That' what I get for doing a code review right after getting up from a nap.

Issue #268, opened 2025-12-02
Refactor and clean up map_shp.c


tvrusso:
There are some open bugs related to shapefile rendering (c.f. #157 , #154 , maybe others) that have gone unaddressed for some time.

Before embarking on that work, it would be a good thing to clean up and refactor the existing `map_shp.c` code for maintainability and readability.

I am not exactly sure where I'm going to start on this, so this issue is likely going to be a parent issue for smaller refactor efforts.

The first focus of this cleanup/refactor will be `draw_shapefile_map`, which is 2673 lines, includes many, many lines of comments, raw X calls, and in some cases duplicated code in case statements to handle various special cases.  Some of the open issues specifically require work on those special cases that could very well break the code as it is and be hard to test.  Breaking this enormous function into small pieces, especially if they can be unit tested, would be a tremendous help in that effort.

Possible early efforts:
   - clean up and remove unnecessary global variables (for example, there's a global variable "npoints" that appears never to be set anywhere, a holdover from when maps.c was split up into many different map rendering files back in commit 2bcc1e9 )
   - reduce overuse of function-scoped variables with a single-character name(this makes it very hard to search for when those variables are used, and reusing the temporary variable "i" for many different things in one function can make it really, really hard to read short sections of code --- for example, there are blocks of code that reference "npoints" in error messages, but actually use the variable "i" for the thing being reported, and the actual setting of "i" is so far removed from the error message lines that it was hard to see what was going on).
   - remove obsolete comments (e.g. there are extensive comments for `draw_shapefile_map` that were completely obsolete when the old hard-coded shapefile rendering was removed, and were obsolescent when dbfawk was first added in 2003)
   - pull code that is cut/pasted many times across multiple `map_*.c` files (such as code that sets a short filename) into reusable, unit-tested functions.
   - pull dbfawk-related search code into small, testable units instead of peppering `draw_shapefile_map` with big blocks of detailed dbfawk calls.  
   - modularize vector drawing actions into reusable chunks that abstract away the raw X calls.
   - Create map unit tests that stub out or create mock, non-graphic versions of the abstract drawing calls.  These tests would assure correct unit conversion, clipping of lines, label placement calculations, ring direction computations, etc. and prevent breakage as deeper hacking proceeds.

[...]

PR #270, opened 2025-12-02
Dbfawk cleanup

### LINKED ISSUE #269: Extract DBFAWK-related code from draw_shapefile_map
When initially written, `draw_shapefile_map` was very specially coded for every single different type of shapefile maps that were supported, with extensive code that used to look at the shapefile name and decide how to render it.  It was, um, difficult to follow.

As long ago as 2003, Alan Crosswell began creating the dbfawk system to separate decisions about how to render different types of shapefiles out of `map_shp.c` and into awk-like files that could be crafted by anyone to render just about any type of shapefile they came across.  It is extremely versatile, and has been the default ever since 2004 (back then it was only disabled if pcre wasn't available or if the user explicitly disabled it), and in 2019 we finally ripped out the non-dbfawk rendering code.

But `draw_shapefile_map` still has large blocks of code to call the various dbfawk functions and set the rendering variables that are really better extracted into separate functions that will make the top level function more readable.

There are also a few common patterns that are cut and pasted in several places across `draw_shapefile_map`.  These are frequently things that free up memory created by dbfawk calls when the function exits (either normally or abnormally), such as:
```
    dbfawk_free_info(fld_info);
    if (sig_info != NULL && sig_info != dbfawk_default_sig  && (sig_info->sig == NULL))
    {
      dbfawk_free_sigs(sig_info);
    }
```
which appears eight times in the function.

There are other issues preventing immediate migration of dbfawk activities into separate functions, such as a reliance on function-local static variables (they must be static because their pointers are stored in a symbol table) --- migrating those to file-local statics would preserve the required permanence of pointers, but allow them to be shared by multiple functions and improve readability of the massive `draw_shapefile_map` code.  

Other issues related to how dbfawk is called will certainly come up and just be taken care of as part of this sub-issue.  I'm already thinking of ways to encapsulate the dbfawk-related variables more cleanly than is currently handled.

There will be no actual behavioral change for this issue, just code consolidation and comment cleanup.

Hopefully, I will be able to get the new functions under unit testing while I'm doing this.

tvrusso:
This PR addresses issue #269, a sub-issue of  #268.

The intent is to move code around `map_shp.c` so that `draw_shapefile_map` is more readable.  

This pr only addresses moving around code that pertains to dbfawk processing in `draw_shapefile_map`.  There may be quite a few commits by the time I'm done, because I'm trying to move very slowly and make commits "atomic".

The second commit in this PR is actually a bug fix unearthed during the attempt to clean up code.

Closes #269

### DIFF
--- a/src/map_shp.c
+++ b/src/map_shp.c
@@ -546,153 +546,31 @@ int shape_ring_direction ( SHPObject *psObject, int Ring )
  * extents) instead of drawing it.
  *
  * The current implementation can draw Polygon, PolyLine, and Point
- * Shapefiles, but only from a few sources (NOAA, Mapshots.com, and
- * ESRI/GeographyNetwork.com).  We don't handle some of the more
- * esoteric formats.  We now handle the "hole" drawing in polygon
- * shapefiles, where one direction around the ring means a fill, and
- * the other direction means a hole in the polygon.
+ * Shapefiles.  Rendering rules for any shapefile are provided by
+ * an associated "awk-like" DBFAWK file.
  *
- * Note that we must currently hard-code the file-recognition
- * portion and the file-drawing portion, because every new source of
- * Shapefiles has a different format, and the fields and field
- * definitions can all change between them.
+ * We handle the "hole" drawing in polygon shapefiles, where one
+ * direction around the ring means a fill, and the other direction
+ * means a hole in the polygon.
  *
  * If alert is NULL, draw every shape that fits the screen.  If
  * non-NULL, draw only the shape that matches the zone number.
  *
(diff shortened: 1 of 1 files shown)


tvrusso:
Of course, as soon as declaring this no longer a WIP I realized there was a huge block of obsolete comments that needed attention, and that's what my last commit did.  It was a comment-change only commit.

I should note that that huge block of comments has been there since 2002 and referenced shapefiles from 2001.  It was clearly just notes from the developer who was still working out how to handle weather alerts.  We now have it pretty well in hand, and it's entirely handled by dbfawk files.  The modified comments reflect the new reality.

Issue #128, opened 2019-05-16
Make dbfawk the only way to render shapefiles


tvrusso:
Currently, dbfawk and the pcre library it requires are separate configuration options (this is a mistake, actually, as requesting dbfawk *requires* pcre, and pcre is used for nothing else --- so it should always have been just ONE option).  When enabled (the default), shapefiles are rendered by first checking if a "dbfawk" file exists to describe how to render the data, and if so, using it to set various rendering options like color, line width, line style, fill style, label size, and even what zoom level at which the feature should no longer be rendered.  If no dbfawk is found, a default (ugly) rendering style is used.

Without dbfawk, shapefile rendering is done by a hard-coded rendering choices that in many cases are based on the file name of the shapefile.

As a result, map_shp.c is peppered with "#ifdef WITH_DBFAWK" ifdefs, making it very difficult to read and work on without the potential of breaking the code that's ifdefed out and probably rarely getting compiled.

dbfawk has been the default  shapefile rendering method since commit 91d3e468 in 2004, and the proliferation of new ".dbfawk" files in our repo to "pretty up" various special use shapefiles as they get discovered is a testimony to the fact that it is being used and used well.

I propose that we simplify our configure process and simplify map_shp.c by making dbfawk the *only* shapefile rendering method.  This requires that we make pcre a mandatory dependency if shapelib is enabled, removing all the ifdef'd code that would be compiled if dbfawk were deselected, removing the ifdefs on "WITH_DBFAWK", and removing the configure option.

Pros:  Simpler, more maintainable code; fewer unnecessary configure choices
Cons: makes pcre mandatory instead of optional for shapefile rendering.

Not suggesting that this be done right away, and certainly not until long after our next release (2.1.2?  2.2.0?)

we7u:
+1

tvrusso:
Jason --- I'm ready to make a pull request on this, so I've deassigned you and reassigned me.  I'm writing the PR now.

PR #272, opened 2025-12-03
Quad overlay flag removal

### LINKED ISSUE #271: Remove "quad_overlay_flag" code from draw_shapefile_map
A very long time ago, there was hard-coded rendering of a special shapefile of USGS 7.5 minute quads (1:24000 scale), named "24kgrid.shp".   When the code saw that it was trying to display a file of this name, it set a flag "quad_overlay_flag" and used this in several conditionals.

The code is further peppered with comments such as:
```
          /* XXX - figure out how to set this from dbfawk: */
```

that have been there since 2003.

Thing is, it was figured out exactly how to do this with dbfawk:  just create a dbfawk file that matches the 24kgrid.dbf signature, and use it to dictate how to render the shapefile.  I committed that dbfawk file to Xastir's repository in 2004.

The file-global variable `quad_overlay_flag` is set to zero at the top of the file and never changed in any way across all of Xastir.   Thus, all the code inside those conditionals is dead code, and only the "else" clauses ever get executed.

Time to remove that stuff and clean up the code.

tvrusso:
Remove vestiges of old hard-coded handling of rendering for the old "24kgrid.shp" shapefile of USGS quad outlines.

These should have been removed in commit 3a9b6e1, but that commit was the result of a use of the "unifdef" utility that stripped out code inside an ifdef, and neglected to remove code that wasn't ifdeffed and just wouldn't get executed once the ifdef was removed.  The blocks removed here are all conditionals that will never have their condition met because the code that set `quad_overlay_flag` to anything but zero was inside the ifdefs that got removed.

Anyhow, this cleans up `draw_shapefile_map` a little and completes work that should have been done in #128.

Closes #271

### DIFF
--- a/src/map_shp.c
+++ b/src/map_shp.c
@@ -2950,9 +2950,6 @@ void draw_shapefile_map (Widget w,
                 }
 
                 (void)XSetForeground(XtDisplay(w), gc, colors[color]); // border color
-
-                // Draw a thicker border for city boundaries
-                (void)XSetForeground(XtDisplay(w), gc, colors[color]); // border
                 (void)XSetFillStyle(XtDisplay(w), gc, FillSolid);
 
                 (void)XDrawLines(XtDisplay(w),


tvrusso:
Converting to a draft because it is currently based on a branch other than master, and will be rebased after #270 is merged.

Issue #288, opened 2025-12-09
Check and possibly fix polygon hole clipping of "wacko" points


tvrusso:
`draw_shapefile_map` has a block of code that addresses polygon shapes with holes by creating an X region, then finding the union of that region with the bounding box of the entire shape, then subtracting from that region each hole polygon in the shape.  Once that is complete, the region is set as the graphics context's clipping mask and all the non-hole polygons are drawn filled.

There is some code in this block that looks questionable:
```
                  if (ok)
                  {
                    // Here we check for really wacko points that will
                    // cause problems with the X drawing routines, and
                    // fix them.  Increment on_screen if any of the
                    // points might be on screen.
                    if (x >  1700l)
                    {
                      x =  1700l;
                    }
                    else if (x < 0l)
                    {
                      x = 0l;
                    }
                    else
                    {
                      on_screen++;
                    }

                    if (y >  1700l)
                    {
                      y =  1700l;
                    }
                    else if (y < 0l)
                    {
                      y = 0l;
                    }
                    else
                    {
                      on_screen++;
                    }
```

The use of hard-coded values for the max screen coordinate seems to suggest that it might not be correct if the screen is actually larger than 1700 pixels in any dimension.

I am NOT addressing this as part of my #283 refactor --- I'm just going to move this code out of the main function and into a helper and use it as is.

However, after #283 is done I need to create a simple polygon shapefile with several holes and do a deep dive into whether Xastir is doing exactly the right thing with it when my window size is large (I've got a 1920x1080 monitor (not even especially large by today's standards), and often have a window that spans the entire width).  I'll have to test all manner of zoom levels so that different parts of the polygon are on or off screen and make sure the rendering is right in all conditions.

It may be that this code is a relic of the era when it was written, and very large monitors weren't common.  It *might* be more correct to use the actual screen max size in the X and Y direction than to use this.  Or this stuff might be exactly correct, I can't tell.

The code that does this check for "wacko" points was appears to have been copied (complete with comment about the point values being "wacko") and altered from other "wacko" point clipping code that was present before Xastir was ever imported into CVS in 2002.  But those older places clipped against 15000l.   The specific checks in the polygon hole code were put in at the time polygon holes were implemented, in commit c4bc91c.  There are no comments indicating why 1700l was picked or what sort of "problems with the X drawing routines" screen coordinate values outside this range might have caused.

[...]

Issue #157, opened 2019-08-06
shapefile polygon labels can be located poorly


kg4wsv:
Shapefile labels for polygons can appear nowhere near the object in question. The existing algorithm uses the center of the bounding box for the polygon.  This results in, for example, the label for "United States" being placed over France due to Alaska being on the other side of xastir's world.

This is mentioned in the code (line 4362 of map_shp.c).  I'm opening an issue because it will become more of an issue as more shape files are available and so there's a git issue associated with the problem.

tvrusso:
On Mon, Aug 05, 2019 at 08:00:57PM -0700, we recorded a bogon-computron collision of the <notifications@github.com> flavor, containing:
> Shapefile labels for polygons can appear nowhere near the object in question. The existing algorithm uses the center of the bounding box for the polygon.  This results in, for example, the label for "United States" being placed over France due to Alaska being on the other side of xastir's world.

My guess is that this shape is a multipart polygon shape, and 
this is just more evidence of Xastir not correctly dealing with those.

There is some work done in the map_shp.c to deal with the fact that there
are multiple "rings" in some polygons, with clockwise coordinates denoting
polygons to fill, and counterclockwise represent "holes" inside those.
Each part of a multipart polygon shape *is* drawn separately and not connected
like multipart polylines are, but only a single label is created for the whole
shape.

It might be possible to base the label location on the centroid of the
"fill" rings rather than the entire shape's bounding box --- putting one 
label in each fill ring rather than one in the entire shape, which, as you say,
is Just Wrong for a multipart polygon.

> This is mentioned in the code (line 4362 of map_shp.c).  I'm opening an issue because it will become more of an issue as more shape files are available and so there's a git issue associated with the problem.

There is no longer a line 4362 of map_shp.c after merge of PR #155 (removal
of non-dbfawk code from Xastir).  I can't find any obviously relevant comment
at line 4362 of the version that existed prior to that merge, either.
Which comment are you referring to here?

kg4wsv:
On Mon, Aug 5, 2019 at 10:23 PM Tom Russo <notifications@github.com> wrote:

> This is mentioned in the code (line 4362 of map_shp.c). I'm opening an
issue because it will become more of an issue as more shape files are
available and so there's a git issue associated with the problem.

There is no longer a line 4362 of map_shp.c after merge of PR #155 (removal
of non-dbfawk code from Xastir). I can't find any obviously relevant comment
at line 4362 of the version that existed prior to that merge, either.
[...]

Issue #154, opened 2019-08-02
Xastir improperly renders multi-part shapefile polylines as a single polyline


tvrusso:
During investigation of some interesting shapefiles for bug #152, we discovered that Xastir is rendering certain shapes in a shapefile improperly.

These shapes are of polyline (SHPT_ARC or SHPT_ARCZ) with more than one "part".  According to the shapelib API at http://shapelib.maptools.org/shp_api.html, a shape of ARC, POLYGON, or MULTIPOINT type may have parts.  Xastir does not support MULTIPOINT, but we do support the other two.  Multiple parts in a shape is indicated by a non-zero value of nParts in the SHPObject data structure returned by SHPReadObject function.

The code in map_shp.c for rendering SHPT_POLYGON polygons is in fact already careful to handle the case when nParts is nonzero, but the SHPT_ARC section of the code blindly constructs a single polyline out of the nVertices points it has in the shape, with no regard to the parts.

This code should be restructured so it looks at nParts, and properly loops over the parts constructing separate, unconnected polylines from them rather than treating the entire list of vertices as a single part.

Work on this would be much easier if issue #128 were closed first, because that would greatly simplify the map_shp.c file, making it much easier to read and modify.

halmueller:
As a workaround, in case you have multipart Shapefiles that you need to use now:
`ogr2ogr -f "ESRI Shapefile" -nlt LINESTRING -explodecollections -lco ENCODING=UTF-8 output.shp input.shp
`
(from https://gis.stackexchange.com/a/226616/16406)

tvrusso:
I should point out that the workaround suggested by @halmueller is probably a desirable thing to do even after this issue is corrected, because Xastir's shapefile renderer also tries to build an rtree spatial index out of shapes, and it is doing so based on the bounding box returned in the min/max coordinates of the object returned by SHPReadObject --- which doesn't tell anything about the multiple parts.   It then uses that rtree to determine which shapes to read from the shape file based on whether the bounding box of the shape intersects the screen.

this means a truly enormous shape (such as one that has one part in one continent, another part half-way across the world) will never really be pruned from consideration.

The breaking up of the multipart shapes using ogr2ogr will work around THAT problem, too.

godfreja:
I think I will take a look at this one. Not an area of the code I am familiar with, but I believe I have an idea on what needs to be done.

godfreja:
Not sure when I will get to this so un-assigning from myself for now.

PR #289, opened 2025-12-09
Refactor SHPT_POLYGON code

### LINKED ISSUE #283: Refactor code in "SHPT_POLYGON" part of the large case statement
This is the big one that makes #274 hard.

This particular case does a lot of hard work, because polygons in shapefiles can actually have multiple parts, and some of those parts might actually represent "holes" in the larger polygon.  That is determined by the direction (clockwise or counterclockwise) of the vertices in the ring.

This thing works by creating a big clip mask in a temporary graphics context that starts with a rectangle that encloses the whole shape object, then cutting out pieces for each hole.  Finally, it draws the filled polygon.

But not all polygons are filled, even those with holes, so it jumps through a lot of hoops trying to figure out what to do, all of it coded right there in the case statement.  This single case of the big switch statement is 909 lines of map_shp.c, making it a LOT to wade through and work out.

tvrusso:
This PR is aimed at #283.

This is going to be a very large PR, with lots of commits, as the code is enormous with lots of special cases, and is easily broken by running through the code with scissors.  I am trying to move slowly and test everything I do at each commit.

The end product should be a set of smaller functions, each called by `draw_shapefile_map`, but that are a bit easier to maintain.

So far, I've done very little but pull big chunks of code into functions, with only minor clean up, but I'm not done yet.

Closes #283 when it's finished.

### DIFF
--- a/src/map_shp.c
+++ b/src/map_shp.c
@@ -966,10 +966,6 @@ void draw_shapefile_map (Widget w,
   // be modified for particular maps in later code.
   (void)XSetLineAttributes(XtDisplay(w), gc, 0, LineSolid, CapButt,JoinMiter);
 
-
-  // NOTE: Setting the color here and in the "else" may not stick if we do more
-  //       complex drawing further down like a SteelBlue lake with a black boundary,
-  //       or if we have labels turned on which resets our color to black.
   if (weather_alert_flag)
   {
     char xbm_path[MAX_FILENAME];
@@ -978,8 +974,7 @@ void draw_shapefile_map (Widget w,
     int ret_val;
 
     // This GC is used for weather alerts (writing to the
-    // pixmap: pixmap_alerts) and _was_ used for beam_heading
-    // rays, but no longer is.
+    // pixmap: pixmap_alerts)
     (void)XSetForeground (XtDisplay (w), gc_tint, colors[(int)alert_color]);
 
     // GXcopy used here because we have been using stippling for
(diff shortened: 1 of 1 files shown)


tvrusso:
This is a big PR with a lot of commits, but it isn't actually that big a deal.

Most of the commits just pull blocks out of `draw_shapefile_map` and into new functions that are just cuts of the code that was pulled out.  

Some repetitive code has been replaced by calls to functions.

A few obvious holes in the code have been corrected (c.f. commits fe8c9ee, 65f2c84, ad79719, and 9d624de (this one actually fixing a dumb thing I'd done in a previous PR)).

The `SHPT_POLYGON` case is now a mere 176 lines, where it used to be 909 lines.  The flow of the function reads more easily.  The blocks are relatively small and might be more easily worked on should work be needed (e.g. for #157 
and #288).

I'm going to declare this refactor done.  Any further breaking up of the code for specific development reasons can be the subject of a future refactor.

In the end, I didn't create any unit tests.  It did not seem like anything here was broken up into anything small enough or separate enough from X to be unit testable, as most of the testing required actually looking at how Xastir rendered things.

I would love to break this up more to the point of being able to unit test the pieces, but I'm running out of steam, and the three big issues that I really wanted this refactor for (#154, #157, and #288) are now something I can get a handle on more easily than before this refactor.

godfreja (approved):

Issue #285, opened 2025-12-07
README.MAPS is grossly outdated and is in need of a rewrite


tvrusso:
README.MAPS is less "documentation" and more "a long discourse on the topic of maps in Xastir."

Much of what it talks about is outdated (e.g. referring to PocketAPRS maps, a feature removed from Xastir in 2019, advocating using shpproj or shpcs2cs to reproject shapefile maps that aren't in lat/lon (the first of which no longer comes with shapelib and the second of which doesn't work with libproj after the 6.0 version), and so forth).  It's got a lot URLs to map resources that are almost certainly dead links.  And some of the links it has are to live links that haven't been updated since 1996.  It recommends WORLDHI.MAP as a good, basic map to start with (c.f. #152)!

It is a mix of useful stuff for new users, highly detailed stuff ranging from basic instruction on how to use unix permissions down to nitty gritty details about how dbfawk works.  It has details about how to make your own maps.  It has details about how to use USGS topo maps in GeoTIFF format (a format abandoned by USGS quite a number of years ago, in favor of GeoPDF).  It even has rants about how some of those GeoTIFF files have been mutilated by non-USGS GIS departments into something unusable, or compressed with a proprietary format.

It even has a discourse on how Xastir computes distances on maps, something no actual user probably needs to know, but developers *might*.

It goes on for almost 1500 lines, and isn't at all something a new user will want to wade through.

A fresh start should be made to replace README.MAPS with proper documentation for users.  That documentation probably belongs on the Xastir Wiki, but that wiki hasn't had a whole lot of love lately and probably needs a lot of work itself.  At a minimum, a (brief!) introduction to the various types of maps we support and how to use them should be readily available with the source code, and shouldn't be so stale that it's practically useless (as it is now).  Recommending online maps should be the default for new users, and for SAR, EMCOMM, ARES, and hams supporting event communication who need offline mapping we should provide a little guidance, but not force them to drink from the firehose.

tvrusso:
As has been pointed out, this file doesn't really document dbfawk all that well, and continues to refer to the non-dbfawk rendering of files (if only to point out that dbfawk wasn't always the way we did business, but it *IS* now, so we needn't even refer to it).

Our documentation is a bit fragmented, with some being on the wiki, some being in these READMES, and some not existing at all.

tvrusso:
[...]

PR #293, opened 2025-12-11
Add a new polygon label location selection method

### LINKED ISSUE #291: Implement simple polygon label point selection from dbfawk
Of immediate concern is getting the large 'countries' shapefile from natural-earth-vector working in Xastir.  This file, and other natural-earth-vector shapefiles just happen to have DBF fields for "label_x" and "label_y" containing the recommended lat/lon for label placement of each shape.

To get these up and running quickly, I would like to implement new dbfawk options such as "label_method", "label_lon", and "label_lat".   "label_method=0" would mean "select label point from center of shape's bounding box", and perhaps "label_method=1" would mean "take label location from label_lon and label_lat" if they're set (and fall back to the bounding box if a shape has blanks for those fields).

This is a hack, but takes advantage of the work that the natural-earth-vector project has done to help out mapmakers by providing that information in the shapefile so the consumer has less work to do.

tvrusso:
This commit adds three new dbfawk variables:
  - label_method
  - label_lon
  - label_lat

If label_method is set in the dbfawk file to 1 and a shape has set label_lon and label_lat to something nonzero, Xastir will now use label_lat and label_lon as the point at which to draw a polygon's label.

If label_method is anything other than 1, or if  (lat,lon) is exactly (0.0,0.0) , we use the label location selection method we've always used, the center of the shape's bounding box as stored in the shapefile object.

Closes #291.

### DIFF
--- a/src/map_shp.c
+++ b/src/map_shp.c
@@ -584,6 +584,9 @@ static int      fill_style,fill_color;
 static int      fill_stipple;
 static int label_color = 8;
 static int font_size = FONT_DEFAULT;
+static int label_method=0;
+static double label_lon=0.0;
+static double label_lat=0.0;
 
 /* default dbfawk rule when no better signature match is found */
 static awk_rule dbfawk_default_rules[] =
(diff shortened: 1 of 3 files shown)


tvrusso:
Images of what the current map looks like with this feature added are in the issue report for the "Built-in map is 35 years old" link, above.

tvrusso:
> We should document the new label_method field values in something user facing (and in repo). However a very quick scan through README.MAPS indicates we currently don't document the colors, so maybe that is something for issue #285.

DBFAWK is primarily documented on the wiki, at https://xastir.org/index.php/HowTo:DBFAWK

I added documentation of the new variables to that.  I also changed the table of colors so that it actually displays the color instead of just its name in xastir.rgb.

Issue #223, opened 2024-08-02
Remove TIGER/Line georeferencing


godfreja:
In a comment in issue #35 @tvrusso said:

> As an aside, there are other ancient, broken features that probably should get removed, such as the use of the geocoder that once enabled us to process TIGER/line data to create an address database with georeferencing, so you could look up street addresses and have Xastir center on them. Those TIGER/Line files are long obsolete, and the [README.md](https://readme.md/) file even points users at an FTP site for downloading preprocessed geocoder files that is long gone. Another issue should probably be opened to call for clearing that out.

This issue is to track that.

tvrusso:
Just for reference:

Address geocoding support was added to Xastir in commit 5f77d06 and subsequent tweaks.

The files that implement the reading of the geocoding database are src/geo-client.c, src/geo-find.c, src/geo.h, src/io-common.c, src/io-mmap.c, and src/io.h.

No code in Xastir actually *creates* the geocoding database, that has to be done by an external program that the commit messages never reference, but which won a Google programming contest (the first one they ran) back around the time this feature was added to Xastir.   (c.f. https://www.cnet.com/tech/services-and-software/google-takes-top-prize-in-its-own-contest/)  It is *possible* that the code could be found in some historical Google page about winning projects, but even if it were, the original TIGER/Line data (in its old topological form, which is wholly unlike the later shapefile format) that it required to work has been discontinued since 2006.  Those files can still be obtained at the Census but would be woefully outdated and useless.

Some Xastir users had gone through the exercise of processing all of the data into useful geocode files and put them on a server, but that site has been down for many years.

I actually still have that file for my state, and just checked --- it still works in Xastir and correctly locates addresses that existed at the time the files were current.  I rather doubt that any current users of Xastir still have those files or use the feature.

tvrusso:
Additional tendrils of this feature work their way into db.c, which uses geo.h and extern data defined therein to mark a found destination on the map.  c.f. commit 4858f0f and 0bac4d0, and other commits to db.c that use `destination_coord_lon`, `destination_coord_lat` and other variables in geo.h.

godfreja:
Thinking out loud here, I wonder about the possibility of updating to make use of Open Street Map data. Either by querying an online source like [Nominatim](https://nominatim.org/) or by working with local OSM data..

Local would be my first choice since it avoid internet dependency, but online query would be easier, both for us to code and for users. (If we only supported online but made the server name configurable an industrious user could locally run Nominatim.)

[...]

PR #301, opened 2025-12-14
Replace the outdated TIGER/Line address lookup with Nominatim based lookup

godfreja:
This PR replaced the geocoding lookup using processed TIGER/Line data with lookups from OpenStreetMaps using Nominatim. The TIGER/Line method relies on out of date data. 

The Maps->Find Address menu item has been replaced with a Maps->Search Location item and adds a Maps->Configure->Geocoding Settings menu. 
The search is free form and supports a filter based on the country.

It should be possible to run Nominatim locally but I have not tried that.

This feature depends on libcurl and the cJSON library. Configure has been updated to account for this.

This is a draft PR as I still need to remove more of the old TIGER/Line lookup code.

### DIFF
--- a/src/db_gui.c
+++ b/src/db_gui.c
@@ -1881,7 +1881,7 @@ void display_file(Widget w)
   Draw_All_CAD_Objects(w);        // Draw all CAD objects, duh.
 
   // Check if we should mark where we found an address
-  if (mark_destination && show_destination_mark)
+  if (mark_destination)
   {
     int offset;
 
--- a/src/xa_config.c
+++ b/src/xa_config.c
@@ -643,7 +643,6 @@ void save_data(void)
     store_string (fout, "SOUND_DIR", SOUND_DIR);
     store_string (fout, "GROUP_DATA_FILE", group_data_file);
     store_string (fout, "GNIS_FILE", locate_gnis_filename);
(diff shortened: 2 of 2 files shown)


tvrusso:
I pulled this branch and built it.  I can search for addresses and it finds them, and the GoTo button works.  But the Mark button seems to do the same thing as the GoTo button, and I don't get a marker on the screen.
nominatim

Is it supposed to mark with a big blue X like you showed in #223?

tvrusso:
Yes, indeed it is.   I had thought maybe that was it.  I hadn't actually looked at code before I posted, but I have since.  That's a relic of the old geocoder thing, iddnit?  And now there's nothing to set or unset it?

I'd try again after deleting that, but I deleted your branch out of my repo and now it's late.  Will try it tomorrow after removing that config line.

tvrusso:
Resurrected your branch using a SHA from my reflog and rebuilt it, after using the older build to reset SHOW_FIND_TARGET to 1 with the old geocoder (I still have an ancient geocode file that works with it).  That made your new one put a mark where I wanted it.

Might want to take out whatever logic there is that cares about that old `SHOW_FIND_TARGET` variable.

tvrusso:
One thing I did NOT test out explicitly was what happens if you try to build when you don't have the required dependencies.  But I see from the basic CI build that that is handled cleanly.

PR #302, opened 2025-12-14
Add dbfawk files for 2025 edition Tiger/Line shapefiles

### LINKED ISSUE #298: Create new dbfawk files for recent TIGER/LINE shapefiles.
The last time we created dbfawk files for TIGER/LINE data from the US Census, it was 2010, and it was for the 2009 edition of TIGER/LINE shapefiles.

None of those dbfawk files work for the 2025 data.

Time to sit down and make new ones.

At a minimum, we should have dbfawks to handle road data, counties, urban areas, county subdivisions, linear and area hydrography, landmarks.

Having a dbfawk for *all* the data types might be asking a lot.

The dbfawks should distinguish features using the MAF/TIGER Feature Class Code (MTFCC) that every one of the files use to define the feature type, and which are documented [here](https://www2.census.gov/geo/pdfs/maps-data/data/tiger/tgrshp2025/TGRSHP2025_TechDoc_E.pdf)

Direct download of the shapefiles can be done [here](https://www.census.gov/cgi-bin/geo/shapefiles/index.php).

tvrusso:
The last time we constructed dbfawks for TIGER/Line shapefiles was in 2010, for the 2009 edition files.  Those dbfawks will not match the signatures of the new 2025 edition files (and probably many other editions in between).

This PR adds new dbfawks for this year's (and maybe slightly older) shapefiles.

Closes #298

### DIFF
--- a/config/Makefile.am
+++ b/config/Makefile.am
@@ -97,6 +97,7 @@ EXTRA_DIST = \
     tl_2025_county.dbfawk \
     tl_2025_cousub.dbfawk \
     tl_2025_linearwater.dbfawk \
+    tl_2025_place.dbfawk \
     tl_2025_pointlm.dbfawk \
     tl_2025_roads.dbfawk \
     tl_2025_rails.dbfawk \
--- a/config/tl_2025_place.dbfawk
+++ b/config/tl_2025_place.dbfawk
@@ -1,0 +1,31 @@
+# Copyright (C) 2000-2023 The Xastir Group
+#
+#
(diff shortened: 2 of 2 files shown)


tvrusso:
At the moment, this branch has been rebased to that of #223, so it appears to have all of those commits in it.  I'll rebase it back onto master and force-push the rebased branch before I'm finished.  If #223 gets closed first, great, otherwise I'll rebase it back onto the master branch without those commits.  I just like the current state of #223 and don't want to have to keep rebuilding Xastir without it.

PR #322, opened 2025-12-26
Clean out nws*.dbfawks

### LINKED ISSUE #303: clean out ancient NWS dbfawk files
Back when dbfawk was new and just getting applied to NWS shapefiles for weather alerts, we found that NWS was changing their DBF fields very often, more often than some Xastir users wanted to download new shapefiles.   Internet was slow and expensive, disk space was precious, and the juice of getting new downloaded weather shapefiles was not often worth the squeeze.

So to address this, every time NWS changed their shapefile's attribute tables, we created a new dbfawk file to go with them, and kept the old one.  That way, people with old collections weren't forced to download new ones, which might be very time-consuming.

We've got dbfawk files for every version of NWS shapefiles that have been released since 2006.  This madness must end.

Dropping all but the latest of each NWS file seems reasonable.   Right now, that's `nwsw_ddmmyy_22.dbfawk`, `nwsfz_ddmmyy_19.dbfawk`, `nwz_ddmmyy_19.dbfawk`, `nwsmzddmmyy_17.dbfawk`, `nwsc_ddmmyy_15.dbfawk`, `nwsozddmmyy_14.dbfawk`,  and maybe a couple of others. Going by the dates on these (the `_XX` generally indicates the year they were created) the changing of dbf fields doesn't happen all that often anymore, and these days downloading new ones isn't as burdensome as it was in 2006.

Cleaning house once every 20 years or so shouldn't hurt.

tvrusso:
Our config directory was full of ancient nws*.dbfawk files that have been preserved over the years to maintain backward compatibility with every variant of NWS shapefile that has come over the internet since about 2006.

There were so many of them it was hard to tell which was which, and which was even the one being used by current shapefiles.

I have removed all nws*.dbfawk files that aren't used for the currently available NWS shapefiles, and renamed the ones that remained to be "nws?_ddmmyy.dbfawk" with no implied year of last update.

We were inconsistent about whether we moved an old one to a new name and kept the unadorned name as the current, or if we just created a new one for every change NWS made to shapefile signatures.

We did this because back in the early 2000s it was onerous to download whole new sets of shapefiles (slow internet) and many Xastir users didn't want to do it.  So we kept dbfawks around that would allow these users to keep their old NWS shapefiles.

But having tons of useless old dbfawks in the directory is a headache, and these days there's no real need for us to preserve backward compatibility to 20 year old NWS shapefiles (or even 10 year old shapefiles).

I have tested that every single one of the current crop of NWS shapefiles has a DBFAWK file associated with it, but that there's only one for each type (c, w, z, fz, oz, mz), and they all have a generic name.

Should NWS change shapefile signatures again, we should not get back in the habit of supporting old versions just so folks don't have to bother downloading the current ones.

Closes #303

### DIFF
--- a/.github/workflows/test-build.yml
+++ b/.github/workflows/test-build.yml
@@ -34,7 +34,7 @@ jobs:
     - name: Update Packages
       run: sudo apt-get update
     - name: Install Packages
-      run: sudo apt-get install -y build-essential xorg-dev graphicsmagick libgraphicsmagick1-dev libmotif-dev libcurl4-openssl-dev libpcre2-dev libdb-dev shapelib libshp-dev libxpm-dev libproj-dev festival festival-dev libgeotiff-dev gpsman gpsmanshp libcjson1 libcjson-dev
+      run: sudo apt-get install -y build-essential xorg-dev graphicsmagick libgraphicsmagick1-dev libmotif-dev libcurl4-openssl-dev libpcre2-dev libdb-dev shapelib libshp-dev libxpm-dev libproj-dev festival festival-dev libgeotiff-dev gpsman gpsmanshp libcjson1 libcjson-dev wget
     - name: Bootstrap (with autoreconf)
       run: autoreconf -i
     - name: configure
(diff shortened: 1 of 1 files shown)


tvrusso:
Going to just merge this one, too.

The PR includes a new check in the Full-Ubuntu build that runs get-NWSdata and verifies that the dbfawks we have in the repo go with the data that get-NWSdata downloads.

I have other stuff I want to do and don't want the unmerged PRs to stack up.

PR #338, opened 2026-01-07
Add 2026 geofabrik dbfawks 

### LINKED ISSUE #32: Update Geofabrik dbfawk files to match current shapefiles
In commit dd99425, Curt added a couple of DBFAWK files to support using shapefiles from http://download.geofabrik.de/ in Xastir.

Unfortunately, the shapefiles currently available on that web site do not match the DBFAWK files anymore.  The dbf signature is different, and they have clearly changed how the feature type is stored (it's now in an fclass field instead of a type field)  I have not checked what range of values the fclass field takes.  It may be that the only difference is in the name of the field, but if they reorganized the values of the field, too, then this may take a little more work.

Opening this bug to solicit community assistance in getting this fixed.

tvrusso:
This is an effort to create dbfawk files for current generation shapefiles from download.geofabrik.de

I am using the suggested dbfawk files from @mthassler to help guide me in my work, but am not using their dbfawks directly --- the color scheme suggested was very nice, but I needed to do more work to refine them.  In the end, I've mostly started from our *old* geofabrik dbfawks that @we7u created, but updating them for new signatures and using the color scheme in @mthassler's work.  I had to refine them a lot by modifying display levels, label levels, and patterns to get my map to look reasonable at middling zoom levels.  

So far I've only got dbfawks for roads and waterways, just as we had before I started.  I will add more to this PR before I mark it ready for review.

Closes #32

### DIFF
--- a/config/OSM_geofabrik_2026_roads.dbfawk
+++ b/config/OSM_geofabrik_2026_roads.dbfawk
@@ -79,4 +79,4 @@ BEGIN_RECORD {key=""; lanes=1; color=8; fill_color=11; name=""; filled=0; patter
 /^fclass=raceway/ {lanes=1; color=4; display_level=64; label_level=16; font_size=1; next}
 /^fclass=t/ {lanes=1; color=4; display_level=64; label_level=16; font_size=1; next}
 /^fclass=tr/ {lanes=1; color=4; display_level=64; label_level=16; font_size=1; next}
-
+/^fclass=unknown/ {lanes=1; color=4; display_level=32; label_level=16; font_size=1; pattern=1; next}


tvrusso:
FWIW, here's what the roads and waterways look like in the Albuquerque, NM area, from zoomed all the way out to city level down to street level.
geofabrik_roads_waterways_farout

geofabrik_roads_waterways_zoomin


geofabrik_roads_waterways_wayin

I've included many features (just as the old geofabrik dbfawks did) such as walkways, cycle routes, driveways, and so forth.  I've tried to make them so they only show up when zoomed really far in.  I tried to declutter the map by minimizing the number of labels that show up until you're zoomed pretty far in.

tvrusso:
I am declaring victory on this one.  I could add the "places" and "landuse" files, but the first one is mostly city polygons that would primarily be useful as a pseudo base layer, and not sure how valuable the second one would be.  And I want to move on.  Railroads might be nice, but I'm stopping here.  The original geofabrik dbfawks that @we7u created were just for roads and waterways, and this is already more than that.  And they all work with today's geofabrik shapefiles.

I have dbfawk files for roads, rivers/waterways, natural features and area water features, and buildings.  I posted some screen shots at various zoom levels on the #32 issue, I think they look pretty good.

PR #359, opened 2026-02-11
Update get-NWSdata for new NWS shapefiles

### LINKED ISSUE #349: Update get-NWSdata after 3 March 2026
The NWS will have new shapefiles to use starting 3 March 2026.  These files are already available, but are not valid until 3 March 2026.

Prior to the release of Xastir 2.2.4, but after 3 March 2026, we need to update the get-NWSdata script and make sure that the dbfawk file associated with each shapefile is correct.

Per discussions among the various people with commit access to the repo, we'll do the 2.2.4 release shortly after getting this done.  Any development work left unfinished by mid-March 2026 will be pushed off to the next development cycle.

tvrusso:
NWS has issued new shapefiles for use after 3 March 2026.

This commit updates get-NWSdata to assure we're ready for it.  I have confirmed that NWS did NOT change any shapefile signatures and so this will be good to go on 3 March 2026.

Closes #349

### DIFF
--- a/scripts/get-NWSdata.in
+++ b/scripts/get-NWSdata.in
@@ -42,13 +42,13 @@
 # should START to be used.  Don't just blindly put the newest filename
 # here if that start-date hasn't arrived yet!
 #
-FILE1="w_18mr25"    # GREEN: NWSM Library: County Warning Area Boundaries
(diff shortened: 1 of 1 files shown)


tvrusso:
This PR is marked as a draft because it shouldn't be merged until 3 March 2026, when NWS starts using their new shapefiles.  But it is ready to go on that date.

PR #363, opened 2026-04-22
Add support for antialiasing via Cairo

### LINKED ISSUE #364: Map label text looks jaggy on high-DPI displays — no antialiasing
### Summary                                                                                                                                                                   

Xastir was originally written for low-resolution CRT monitors where text rendering without antialiasing was acceptable and common. On modern high-DPI displays, especially when the user selects larger map label fonts, diagonal strokes in glyphs appear visibly jagged because no antialiasing is applied to map label rendering.                     

### Steps to reproduce                                                                                                                                                        

1. Run Xastir on a high-DPI display (e.g. laptop ≥ 1080p, HiDPI / Retina, or 4K).                                                                                             
2. Increase the map label font size in the preferences.
3. Observe any rotated or horizontal label drawn on the map.                                                                                                                  

### Expected behavior

Label glyphs should render smoothly, with antialiased edges, at any size and rotation — comparable to text rendered by any modern desktop application.                        

### Actual behavior                                                                                                                                                           

Glyph edges are hard-pixel only. Diagonal strokes show staircase aliasing, which becomes progressively worse (more visible) as font size grows.                               

### Root cause                                                                                                                                                                

Map text is drawn through the legacy xvertext path, which uses raw X11 bitmap rendering with no subpixel or antialiasing support.                                             

### Proposed fix                                                                                                                                                              

Render map label text through a modern font renderer that supports antialiasing — **Cairo** (with `cairo-xlib`) is a natural fit since it integrates directly with the        
existing X11 drawable, supports arbitrary rotation, and is already widely available on all target platforms (Linux, *BSD, macOS via XQuartz).

Example of the issue (note blue text around the map particularly the letter W, and station fonts)
Image

hjf:
This PR adds support for antialiasing fonts via Cairo

The fonts only affect the map area (the rest of the UI is still Motif widgets which are pretty difficult to port)

After this change, map coordinates and stations are rendered with much nicer looking text.

Notice text on map borders and station labels.

Screenshot_20260422_154639

These changes were created with AI assistance.

Closes #364


tvrusso:
Making Xastir look nicer is a good thing and I'm all for it, even though its X11/Motif interface is clunky and will probably need to be replaced.   Nothing gets done in Xastir unless someone steps up and offers to do it, so thank you very much for being willing to work on this and offer it to the project.

I am absolutely NOT averse to this change, but am not ready to review or approve it as I'm very busy with other projects at the moment.  Fortunately, you're still working on it so we have time.   

It would be very helpful if, as you refine this PR, you try to do a few things:

  - make sure that the code falls back gracefully to old behavior if the user does not have cairo installed
  - allow a user to explicitly disable probing for and using cairo if they don't want it (using configure options like "--without-cairo")
  - Make sure to output a line of text in the final block of output to show whether or not cairo anti-aliasing support is enabled (just like we do for things like shapelib, geotiff, etc.)
  - Make sure configure continues to work if pkgconfig is not installed (i.e., skip any configure tests that require it and assume the package that would have required it doesn't exist)
  - Please don't replace hard-coded ISO8559-1 characters in comments.  They are outmoded and out of favor, but in comments they are more readable.
  - If any ISO8559-1 characters are hard-coded in print/sprintf/whatever statements, it is in general NOT safe to replace those with UTF-8 unless all the code downstream is adjusted to handle those new characters.  We were once bitten by this and it made all display of the degree sign break in dialog boxes.   Tread carefully and test, retest, and re-retest any such changes.
 - Don't put system-specific cairo/pkgconfig/cairo-xlib install instructions directly into  INSTALL.md, which is intended as generic guidance.  If you notice, all the instructions for other packages simply name the upstream package that provides the feature, not the system-specific package name or combination of packages needed to use it.   System-specific installation guidance is done via system-specific install notes on the wiki.

Please carefully review any code you have had AI generate.  I am also working on a project that has heavy AI involvement and I'm finding it both a wonderful coding partner and a frustrating one at the same time --- it is phenomenally helpful at times, but often produces code that has subtle bugs or misfeatures that I have to tell *it* about and get it to correct, especially when it involves delicate interactions between the new code and old code it doesn't quite understand correctly.

When you're ready for this to be reviewed and approved, Jason and I will go over it as well.

It would be phenomenally helpful if you would open  an issue report associated with this PR, and in that issue report show some before/after screenshots of the improvements provided by this code change.

tvrusso:
One thing I note is that you changed the "basic-ubuntu-build" action to include cairo-dev.  This is the build where we try to have no optional packages installed at all to make sure the minimum build works, so it is more appropriate for this one NOT to have cairo installed to verify that the fallback code builds properly.

The "full-ubuntu-build" is the one where we install *all* optional packages and make sure that builds.

Please return "basic-ubuntu-build" to this minimum configuration, as it provides a valuable check that our so-called minimal install continues to build.

hjf:
@tvrusso I updated the PR and created the ticket.

As for the basic build, it will not fail because it'll fall back to the previous behavior if Cairo isn't found.

In the full build we install libcairo and it'll be enabled by default.

tvrusso:
Yep, saw that.  Thanks for that update.

I still don't have much time to do a full review, but thanks for taking the effort to improve Xastir in this way.  I look forward to checking it out when I come up for air and seeing how much nicer fonts render with these changes.

godfreja:
I gave this a quick try on my Mac. The border text on the map is noticeably better, but the station text is noticeably harder to read with this change. I'll need to investgate- my first guess is some sort of font choice issue.
Screenshot 2026-04-24 at 2 09 55 PM

Screenshot 2026-04-24 at 1 56 07 PM

First picture is the before, second picture is the after.

hjf:
@tvrusso i think it's a combination of font weight (maybe choose a bold font, I just used "Helvetica" that was selected by default) and the very strong contrast between the black border and white text. i'll have to check if the black border could be reduced (since we're doing antialiasing we can do subpixel adjustments). maybe also change the color too so we don't have the grey-black-white transition

tvrusso:
> @tvrusso i think it's a combination of font weight (maybe choose a bold font, I just used "Helvetica" that was selected by default) and the very strong contrast between the black border and white text. i'll have to check if the black border could be reduced (since we're doing antialiasing we can do subpixel adjustments). maybe also change the color too so we don't have the grey-black-white transition

It was actually @godfreja who tried it out and saw that, not me.  I still haven't had a chance to try your modified code.

godfreja:
GH CoPilot suggested [this fix](https://github.com/user-attachments/files/27068685/font-fix.patch) which seemed to work well with the quick try I gave it.
Screenshot 2026-04-24 at 5 11 38 PM

hjf:
@godfreja  I've been trying multiple alternatives, like drop shadows and such. and it seems the maximum legibility is black text on a white outline. It's what OpenStreetMap does on their maps. it looks good both on OSM Mapnik:
image
and on grey vector maps
image

tvrusso:
@godfreja and @hjf  do your different solutions work with all the various station text format options we have in the Map configuration menu?

station_text_menu

hjf:
@tvrusso yes though I asked the AI for an improved drop shadow and it made me one with blur which looks pretty good. here are the different alternatives with the current PR:

Black border:
image
Black shadow:
image
Text on black:
image
Drop shadow:
image

Outside this PR is my favorite, black on thick white:

image

src/maps.c
```diff
@@ -1825,12 +1826,12 @@ void draw_complete_lat_lon_grid(Widget w)
 // is used.
 //
 // For MGRS and UTM-Special grid only:
-// UTM Zone 32 has been widened to 9� (at the expense of zone 31)
-// between latitudes 56� and 64� (band V) to accommodate southwest
-// Norway. Thus zone 32 extends westwards to 3�E in the North Sea.
-// Similarly, between 72� and 84� (band X), zones 33 and 35 have
-// been widened to 12� to accommodate Svalbard. To compensate for
-// these 12� wide zones, zones 31 and 37 are widened to 9� and zones
+// UTM Zone 32 has been widened to 9� (at the expense of zone 31)
```
godfreja: Tom already mentioned this, but symbols have been changed here. (I had this problem just opening a file in VSCode. I had to change the encoding and manually copy and paste symbols back. It was a pain.)

godfreja (commented):
Other than the symbol changes I think the code looks good once an update for the illegible text is checked in. (I will do some more testing once a fix is pushed.)
Thanks for doing this, this is a nice change.

src/maps.c
tvrusso: Yeah, a lot of editors have quietly modified encoding without warning.  I had a really hard time with that when I was fixing up the French translation of the helpfile, which wound up shipping in release 2.2.4 with unreadable UTF-8 characters for every accented letter (because Xastir does not correctly display UTF-8 in any of its dialogs or map window).  I had to use 
```
iconv -i UTF-8 -t ISO-8859-1 help-French.dat > help-French.dat_new
```
to fix it.  That tool might help here, too.

hjf: I've created PR #366 to fix this for good.

tvrusso: That's a fine fix for the language files (THANK YOU!), but the changes we're requesting are in maps.c, where degree signs in comments were switched to UTF-8.   In comments it's not a huge deal, but we'd like that restored to ISO8559-1.

tvrusso (dismissed):
This looks great to me.  I do note that the border appears quite a bit wider than it used to before the change, but that's not enough for me to reject the change.

tvrusso (commented):
On second thought, I don't like the illegibility of the station labels, they appear just as bad for me as they did for Jason.

Jason provided a patch that seems to improve things, I think that should be added to the PR.

tvrusso:
Before Jason's patch:

bad_station_text

After:

nicer_station_text



The patch can be found linked to in this comment:https://github.com/Xastir/Xastir/pull/363#issuecomment-4316722848

hjf:
applied @godfreja's patch

tvrusso (approved):
Now I approve.   Looks great, even the station labels.

tvrusso:
No need to rebase this, I expect that Jason will approve soon and I'll just merge it when he does.

godfreja:
I probably won't be able to test the update until tonight or tomorrow. I've been reviewing code today on my tablet from a parking lot interspersed with testing repeater coverage.

godfreja (approved):

tvrusso:
I not only merged this commit, I edited the "Installing Xastir" wiki page to add "cairo" to the list of optional libraries, and the three system-specific install notes files (FreeBSD, Ubuntu, and MacOS with Homebrew) to include the same packages we install in the CI testing.

Thank you again.

PR #366, opened 2026-04-26
Convert language files to UTF-8, decode to Latin-1 at load time

### LINKED ISSUE #365: Language files break silently when re-saved by modern editors (ISO-8859-1 → UTF-8 corruption)
All config/language-*.sys files are stored in ISO-8859-1. Any modern text editor (VS Code, gedit, nano with locale settings, etc.) will silently re-save them as UTF-8 when a contributor edits a translation. The result is that accented characters — é, ü, ñ, ç, etc. — are stored as 2-byte UTF-8 sequences but Motif/Xlib reads them as single Latin-1 bytes, producing mojibake in every menu and dialog label of the affected language.

Steps to reproduce:

- Open language-French.sys in VS Code (or any UTF-8-defaulting editor).
- Fix a typo, save.
- Build and run Xastir with the French language selected.
- Observe garbled accented characters throughout the UI.

Impact: Any contributor who touches a translation file silently breaks that language for all users. There is no error at load time; the corruption is invisible until runtime.

Proposed fix:

- Convert all language-*.sys files to UTF-8 (one-time iconv pass).
- Add a UTF-8 → ISO-8859-1 translation layer inside load_language_file() in lang.c so the app always loads UTF-8 files and internally keeps feeding Latin-1 bytes to Motif, requiring zero changes anywhere else.

hjf:
### Problem

`config/language-*.sys` were stored in ISO-8859-1. Modern editors default to UTF-8 and will sometimes silently re-encode files on save. AI tools will use blind "replacements" ignoring file encoding. This caused accented characters in French, German, Spanish, Italian, Dutch, and Portuguese translations to be stored as multi-byte UTF-8 sequences while the loader copied the raw bytes straight into Motif widget labels — producing mojibake with no warning.

### Solution

**Two-part, minimal change:**

1. **One-time conversion** — all 12 `language-*.sys` files converted from ISO-8859-1 to UTF-8 via `iconv`. Files are now safe to edit and commit from any modern editor.

2. **Translation layer in `lang.c`** — added `utf8_to_latin1_inplace()`, called on each translated string immediately after it is parsed from the file, before it enters `lang_buffer`. Motif and the rest of the app continue to receive Latin-1 bytes unchanged. No other files were modified.

### Properties of the converter

| Input | Output |
|---|---|
| ASCII (U+0000–U+007F) | copied as-is |
| Latin-1 supplement (U+0080–U+00FF) | decoded from 2-byte UTF-8 → single Latin-1 byte |
| Outside Latin-1 (U+0100+) | replaced with `?` (safe, visible fallback) |
| Invalid UTF-8 bytes | passed through unchanged (legacy file safety) |

### Testing

- `make lang.o` — clean compile, zero warnings.
- `file config/language-*.sys` — all report `Unicode text, UTF-8 text`.
- Verified round-trip: a French string containing `é` (U+00E9) is stored as `0xC3 0xA9` in the file and arrives in `lang_buffer` as `0xE9`, identical to the original ISO-8859-1 encoding.

### Scope

No behaviour changes outside `lang.c`. No new dependencies. No changes to the `.sys` file format — the pipe-delimited structure is identical.

Screenshot of the spanish language translation after this change:

image

fixes #365

### DIFF
--- a/config/language-Dutch.sys
+++ b/config/language-Dutch.sys
@@ -758,7 +758,7 @@ WPUPSTI012|laatste via onbekend||
 WPUPSTI013|, en is van positie veranderd||
 WPUPSTI014|Huidig Vermogen en Winst:||
 WPUPSTI016|Hoogte: %.0f%s ||
-WPUPSTI017|Richting: %sø ||
+WPUPSTI017|Richting: %sÃ¸ ||
 WPUPSTI018|Snelheid: %.1fkm/h||
 WPUPSTI019|Snelheid: %.1fmph||
 WPUPSTI020|%0.1f Mijl||
@@ -770,10 +770,10 @@ WPUPSTI025|Wind Richting: %s Snelheid: %03d km/h||
 WPUPSTI026|Wind Richting: %s Snelheid: %s mph||
 WPUPSTI027| Piek: %03d km/h||
 WPUPSTI028| Piek: %s mph||
-WPUPSTI029|Temperatuur: %02.1føC   ||
-WPUPSTI030|Temperatuur: %søF   ||
+WPUPSTI029|Temperatuur: %02.1fÃ¸C   ||
+WPUPSTI030|Temperatuur: %sÃ¸F   ||
 WPUPSTI031|Vochigheid: %s%%   ||
-WPUPSTI032|Gevoels temperatuur: %02.1føC  ||
+WPUPSTI032|Gevoels temperatuur: %02.1fÃ¸C  ||
 WPUPSTI033|Baro: %s hPa||
 WPUPSTI034|Sneeuw: %0.1f (cm/24h)||
 WPUPSTI035|Sneeuw: %0.0f (inch/24h)||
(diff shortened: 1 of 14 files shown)


tvrusso:
There are more files to worry about than just the config/lang-*.sys.

The big one that screwed up before release 2.2.4 was help/help-French.dat, which got encoded into UTF-8 by a silent editor conversion and then rendered horribly when using "Aide->Index de l'Aide" while the French language is selected.

Can your change be updated to perform the same conversion-on-read for the help files?  (Most are horribly outdated, but I spent days on the French one bringing it up to date, and the UTF-8 mess was embarrassing because it wasn't discovered until after the release).

The help file is read in by the Help_Index function in main.c, and it seems that it might be a simple matter to apply your utf8-to-latin function here as the file is read in.

hjf:
@tvrusso I've had the AI find more cases but it only found the help files. I've updated the PR with a handler for these.

tvrusso:
This is great!   

french_help

I'm going to accept that right off the bat.

THANK YOU!

tvrusso (approved):
Assuring the code is using latin-1 data to feed to Motif by *converting it on read* is the ideal solution to this problem.

PR #367, opened 2026-04-26
Work around ugly UTF-8 map label display

### LINKED ISSUE #343: Xastir can't render map labels from shapefiles with UTF-8-encoded fields
Xastir does a bunch of manipulation of the locale using the "setlocale" function, forcing the locale to "C" in a bunch of places, then never setting it back to the user's default.

This has been getting done since at least 2002, when Xastir was imported into CVS, and numerous attempts have been made to fiddle with the setlocale calls over the years.

This makes Xastir display map labels in UTF-8 form as messy, ugly text:

Image

That should read "Cañon de Carnué Land Grand Hall and Annex", but the UTF-8 is not rendered properly.  This example is right out of OpenStreetMap building shapefiles.

Unfortunately, all that setlocale stuff is there because Xastir is not properly processing things like its config files if the locale is not "C".

Locales were also broken in 2019 when an attempt was made to support the Greek language --- the changes made to support Greek broke display of the degree symbol in every language in things like the station info boxes.  A lot of commits back then tried to fiddle with the calls, inserting some, trying to reset back, and so forth, but it never worked and the Greek language support commits were reverted.

Work should be done to clean up Xastir's processing of its init file so it works no matter *what* the user's locale is set to, and should properly render maps when the shapefile has fields in UTF-8.

This is not going to be simple, will probably require some refactor of the init and config file processing, and will require very careful testing with and without the user's local set to support UTF-8.  We will also have to make sure that Xastir doesn't try to transmit UTF-8 encoded strings.  A lot of unit tests will be required, which means lots of refactoring to make blocks that are unit testable.

More and more map data is coming out with exactly such labels.

tvrusso:
Issue #343 pointed out that many new shapefiles are getting produced with UTF-8 encoded accented characters, and these display as mojibake in Xastir's map window because Motif doesn't grok UTF-8 without a great deal of coding support that we're not really interested in persuing.

A recent change done in PR #366 implemented a simple utf8_to_latin1_inplace function that provides an immediate workaround for this map display issue.

This PR applies that function to all shapefile map labels before they're used for anything (which includes displaying them, but also storing them in or accessing them from hashes for decluttering purposes).

After application, maps with UTF-8 encoded fields used for labeling are now rendered nicely in Xastir.

Closes #343

### DIFF
--- a/src/map_shp.c
+++ b/src/map_shp.c
@@ -82,6 +82,7 @@
 #include "rotated.h"
 #include "color.h"
 #include "xa_config.h"
+#include "lang.h"
 
 #ifdef HAVE_LIBSHP
   #include "awk.h"
@@ -1322,6 +1323,7 @@ void draw_shapefile_map (Widget w,
               // the symbol and aligned nicely.
               if (map_labels && !skip_label)
               {
+                utf8_to_latin1_inplace(temp);
                 draw_nice_string(w, pixmap, 0, x+10, y+5, (char*)temp, 0xf, 0x10, strlen(temp));
               }
             }
@@ -1410,6 +1412,7 @@ void draw_shapefile_map (Widget w,
                    && map_labels
                    && !skip_label )
               {
+                utf8_to_latin1_inplace(temp);
                 x=points[0].x;
                 y=points[0].y;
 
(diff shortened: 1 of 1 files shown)


tvrusso:
I have been reviewing this PR using maps from download.geofabrik.de.   The maps displayed in the issue report are from the New Mexico shapefile set (roads and buildings) and the Brandenburg set (roads and buildings) for Germany.  Both contain UTF-8 characters in some feature names, and were horribly garbled before this PR.

godfreja (approved):
I haven't tested it but the code looks good

PR #370, opened 2026-04-27
Extract popup dialog logic into popup_controller

hjf:
First pass of the *_gui.c controller-extraction effort. Lifts the business logic of the popup-message dialogs out of Motif callbacks into a self-contained controller that can be unit-tested without linking against Xt/Xm.

Adds src/popup_controller.{h,c} (80 + 147 LOC). Retires the file-scope global popup_time_out_check_last by folding it into popup_controller_t.last_timeout_check. The Widget bag (pw[]) and the popup_message_dialog_lock mutex stay in popup_gui.c -- they are Motif/threading concerns out of scope for this pass. Dead code (pwb, written by clear_popup_message_windows but never read) is also left untouched.

Behavior preserved: when all popup slots are full, the new popup is still silently dropped (legacy i=0 fallback kept in popup_message_always). popup_ID_message (rotated ATV ID rendering) is pure X11 graphics and is not extracted.

Tests: tests/test_popup_controller.c (11 cases). The controller has zero Xastir-side dependencies, so the test binary needs no stubs file -- first validation of the "self-contained controller header" rule. Full suite: 241/241 ok.

Note for reviewers: slot occupancy is now tracked in two places (Widget != NULL on the GUI side; controller.slots[i].active in the controller). Kept lockstep at the four mutation sites. Candidate for unification once more dialogs are migrated.

Part of the *_gui.c controller-extraction effort — see umbrella
issue #TBD (replace with the GitHub issue link before opening).
This is the first PR in a 9-PR stack; review independently or
follow the stack as `de-spaghetti-1` … `de-spaghetti-9`.

### Why

The popup-message dialogs in `src/popup_gui.c` mix Motif callback
plumbing with the small amount of state and decision logic that
governs when a popup should appear, which slot it occupies, and
when it should time out. That logic can't currently be unit-tested
because it's locked inside Xt/Xm callbacks. This PR is the first
shakedown of the pattern: lift the logic kernel out into a
self-contained controller so it can be tested directly.

### What changes

- New `src/popup_controller.{h,c}` (80 + 147 LOC). Pure standard C,
  no `<Xm/*.h>` includes, no Xastir-side dependencies.
- `popup_gui.c` callbacks are thinned to widget-marshalling only;
  decisions go through the controller.
- File-scope global `popup_time_out_check_last` retired (folded
  into `popup_controller_t.last_timeout_check`).
- New `tests/test_popup_controller.c` (11 cases) wired into the
  Autotest harness via `tests/popup_controller_tests.at`.
- Full test suite: 241/241 passing.

### What is intentionally not changed

- The Widget bag `pw[MAX_POPUPS]` and `popup_message_dialog_lock`
  mutex stay in `popup_gui.c` (Motif/threading concerns, out of
  scope for the logic split).
- `pwb` (Popup_Window) is dead code — written by
  `clear_popup_message_windows`, never read. Left untouched here.
- `popup_ID_message` (rotated ATV ID rendering) is pure X11
  graphics with no logic kernel worth extracting.

### Behavior preservation

- "All slots full" still silently drops the new popup (the legacy
  `i=0` fallback in `popup_message_always` is preserved).
- All visible popup behavior is unchanged.

### Reviewer hints

- The interesting file is `src/popup_controller.h` — the struct
  defines exactly what state the popup logic owns.
- One known asymmetry: slot occupancy is now tracked in two places
  (`Widget != NULL` on the GUI side, `slots[i].active` in the
  controller). They are kept lockstep at the four mutation sites.
  Candidate for unification once a few more dialogs are migrated.

### Tests

```
./bootstrap.sh && ./configure && make && make check
```

Should report 241/241 ok.

### DIFF
--- a/MIGRATION_PLAN.md
+++ b/MIGRATION_PLAN.md
@@ -1,0 +1,96 @@
+# Xastir Presentation/Logic Separation — Migration Plan
+
+This plan operationalizes Phase 2 of the Qt migration: extract domain logic
+from Motif callbacks into testable controllers, one `*_gui.c` file at a time.
+The Qt port itself is downstream of this work; success here is measured by
+shrinking the global `Widget` table and growing the unit-test suite, not by
+toolkit changes.
+
+## Guiding constraints (carried over from prior decisions)
+
+- **De-spaghetti first.** Each migrated file ends with logic in a controller,
+  not in callbacks. Toolkit swap is a later concern.
+- **C compiled as C++** is the agreed language strategy. New controllers stay
+  in C-style structs and free functions for now — no STL, no RAII rewrites.
+- **Map canvas (`maps.c`, `draw_symbols.c`, `map_*.c`) is out of scope.** Don't
+  refactor map rendering as part of this work.
+- **No new Cairo surface area.** Cairo retires when Qt arrives.
+- **Tests are the safety net.** Every migrated file gets at least one new
+  `test_<feature>_controller.c` covering its extracted logic, wired into the
+  existing Autotest harness in `tests/`.
+
+## The pattern (apply to every file)
+
+For each `src/<feature>_gui.c`:
+
+1. **Inventory.** List every global the file reads or writes — `Widget`
+   handles, dialog flags, current selections, transient buffers — and every
+   non-trivial helper that doesn't directly call Motif. The result is the
+   future controller's state and methods.
+2. **Create `src/<feature>_controller.{h,c}`.** Define a struct that owns the
+   non-Widget state. Move the pure-logic helpers in as functions taking a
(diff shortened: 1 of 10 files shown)


hjf:
@godfreja @tvrusso this is what a PR would look like. i extracted the 9 commits into 9 separate branches, and created commit messages accordingly. so we'll have 9 separate PRs. i won't create the other 8 immediately of course, as it would spam this repo innecessarily

this was created, as you know, with AI assistance.

godfreja:
@hjf - is this something you want feedback on?
This is my busy season, so it may take a little while if so.

hjf:
@godfreja just left this here. I canceled my subscription to the LLM that produced this PR hah.

PR #375, opened 2026-05-05
map_OSM: replace per-pixel X11 calls with XImage bulk transfer

### LINKED ISSUE #374: OpenStreetMaps render slowly
# OSM tiled map rendering is extremely slow, especially on Raspberry Pi 2

## Summary

Zooming in or out on an OSM tiled map takes approximately 2 seconds to redraw on a
modern PC with 32 GB RAM. On a Raspberry Pi 2 the map effectively freezes for many
seconds on each zoom or pan operation, even when all tiles are already present in
the local cache (no network activity involved). Moving the tile cache to a RAM disk
produces no measurable improvement, confirming the bottleneck is not I/O.

## Steps to reproduce

1. Configure an OSM tiled map source (e.g. `OSM_tiled_mapnik.geo`).
2. Allow tiles to download and populate the local cache.
3. Click zoom in or zoom out.
4. Observe the delay before the map redraws.

On a 1920×1080 display at a mid-level zoom the delay is ~2 s on a modern desktop.
On a Raspberry Pi 2 the delay is much longer and the UI becomes unresponsive.

hjf:
draw_OSM_image() was calling XSetForeground() + XFillRectangle() for every individual pixel. On a 1920x1080 display with OSM tiles this amounts to ~2 million X11 round-trips per redraw, which is extremely slow on Raspberry Pi 2 hardware and noticeably slow (~2s) even on modern PCs.

Changes:
- Allocate a client-side XImage (screen-sized) before the pixel loop
- Write pixels into the XImage buffer using XPutPixel()
- Flush the entire buffer to the pixmap with a single XPutImage() call
- On allocation failure, fall back to the original XSetForeground path

Additionally, eliminate the intermediate ImageMagick canvas in draw_OSM_tiles(). Previously each tile was decoded with ReadImage() then composited onto a large in-memory canvas via CompositeImage(), and only then passed to draw_OSM_image(). This wasted memory and CPU proportional to the number of visible tiles.

Now each tile is decoded and rendered directly into the shared XImage using per-tile tiepoints, completely removing the canvas allocation, SetImage(), SetImageOpacity(), and N x CompositeImage() calls.

Tested on x86-64. Subjective improvement from ~2s redraw to near- instant on cached tiles. Expected to be significantly faster on Raspberry Pi 2 where the original bottleneck was most severe.

This PR was made with AI Assistance.

closes #374

### DIFF
--- a/src/map_OSM.c
+++ b/src/map_OSM.c
@@ -719,15 +719,20 @@ static void draw_image(
 
 
 /**********************************************************
- * draw_OSM_image() - copy map image to display
+ * render_OSM_image_pixels() - internal helper: write one decoded image's
+ * pixels into a caller-supplied XImage buffer (no ownership, no XPutImage).
+ * Called by both draw_OSM_image() (single image) and draw_OSM_tiles()
+ * (shared buffer for all tiles) so that the X server round-trips for
+ * XGetImage and XPutImage happen exactly once per redraw, not once per tile.
  **********************************************************/
-static void draw_OSM_image(
+static void render_OSM_image_pixels(
   Widget w,
   Image *image,
   ExceptionInfo *except_ptr,
   tiepoint *tpNW,
   tiepoint *tpSE,
-  int osm_zl)
+  int osm_zl,
+  XImage *ximg)    // XImage to write into; may be NULL (fallback to X11 calls)
 {
   int l;
   XColor my_colors[256];
@@ -745,6 +750,7 @@ static void draw_OSM_image(
   long scr_x,  scr_y;             // screen pixel plot positions
   long scr_xp, scr_yp;            // previous screen plot positions
   int  scr_dx, scr_dy;            // increments in screen plot positions
+  unsigned long fill_pixel = 0;   // pixel value accumulated before writing
 
   //if (debug_level & 512)
   //    fprintf(stderr,"Color depth is %i \n", (int)image->depth);
@@ -930,17 +936,8 @@ static void draw_OSM_image(
     HandlePendingEvents(app_context);
     if (interrupt_drawing_now)
     {
-      // Update to screen
-      (void)XCopyArea(XtDisplay(da),
-                      pixmap,
-                      XtWindow(da),
-                      gc,
-                      0,
-                      0,
-                      (unsigned int)screen_width,
-                      (unsigned int)screen_height,
-                      0,
-                      0);
+      // Caller owns ximg; we just return without flushing —
+      // draw_OSM_image() or draw_OSM_tiles() will handle cleanup.
       return;
     }
 
@@ -997,7 +994,7 @@ static void draw_OSM_image(
               {
                 continue;
               }
-              XSetForeground(XtDisplay(w), gc, my_colors[(int)index_pack[l]].pixel);
+              fill_pixel = my_colors[(int)index_pack[l]].pixel;
             }
             else
             {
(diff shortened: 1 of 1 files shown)


src/map_OSM.c
```diff
@@ -1274,134 +1337,100 @@ void draw_OSM_tiles (Widget w,
     /*
      * Create a canvas upon which the tiles will be composited.
```
tvrusso, line 1332: This comment is now obsolete, as the code it refers to has been deleted.

src/map_OSM.c
```diff
@@ -727,7 +727,8 @@ static void draw_OSM_image(
   ExceptionInfo *except_ptr,
   tiepoint *tpNW,
   tiepoint *tpSE,
-  int osm_zl)
+  int osm_zl,
+  XImage *ext_ximg)   // optional external XImage; if non-NULL, skip alloc and XPutImage
```
tvrusso: What is the reason for this optional external XImage argument?   draw_osm_image is never called with non-null for this argument.   Is there some future improvement that might make it useful?

There are a few places in the code where it's important whether this is given or not.

tvrusso:
I am running this code now and it does seem to work very well.   Just had one comment and one question I'd like resolved before I approve.

tvrusso:
One more comment:   The code in map_OSM.c was designed by Jerry Dunmire and one of the reasons for the canvas compositor was to handle missing tiles, c.f. commit c1df1583 where that code was introduced.   Apparently before that change, missing tiles would wind up as big red blocks, whereas after the canvas code and transparency was put in lower zoom level tiles would show through the missing blocks and at least make a map that was complete, if not nice.

I'm not sure if that's a problem we need to worry about much, but it was why that code was put in and why transparency was being handled the way it was.   I'm not sure that this PR maintains that fallback.

I am also not sure whether it is that important to fallback that way nowadays.

But one of the features of our OSM tile cache is that it can still be used offline.   If one is using it offline and doesn't have some tiles while zooming and panning, it could get ugly.

That should be tested out by going offline and zooming/panning out of the area where you have good tile caches at multiple zooms.

src/map_OSM.c
```diff
@@ -1274,134 +1337,100 @@ void draw_OSM_tiles (Widget w,
     /*
      * Create a canvas upon which the tiles will be composited.
     */
-    canvas_info=CloneImageInfo((ImageInfo *)NULL);
```
tvrusso, line 1277: The deletion of this code (the selected line and all that follows) concerns me, as it was put in for a reason.   I do not know how important that reason is or how much of a concern it remains, but casually discarding it is dangerous and needs more justificiation than an AI claiming it's unnecessary.

Please see commit c1df1583 and those following it for map_OSM.c.

I believe a great deal of testing with missing tiles and offline use will be necessary before I can sign off on this.

hjf:
Updated the code to address all three review threads. Latest commit: bea953fc

**`ext_ximg` parameter (thread 2):** The parameter has been removed entirely. The approach has been refactored: the pixel-writing loop is now in a new internal function `render_OSM_image_pixels()` that takes a plain `XImage *` (no ownership semantics). `draw_OSM_image()` is a thin wrapper that allocates/frees the XImage and calls it. `draw_OSM_tiles()` does the same with a shared XImage across all tiles. This makes ownership unambiguous throughout.

**Missing tiles / offline use (thread 3 and issue comment):** This is now explicitly handled. Both `draw_OSM_tiles()` and `draw_OSM_image()` initialize the XImage buffer using `XGetImage()` against the current pixmap rather than `calloc()`. This means:
- The buffer starts with whatever is already rendered (e.g. lower-zoom tiles from a previous redraw)
- Tiles that are missing (file not found, offline) simply don't overwrite those areas — the previous content shows through, exactly the same fallback behavior that the canvas+transparency approach provided
- There is a `calloc` fallback in case `XGetImage` fails, which preserves the fast path at the cost of showing black for missing tiles in that rare case

The performance model is now: 1× `XGetImage` → N× pixel loops (CPU only) → 1× `XPutImage`, regardless of how many tiles are in view.

hjf:
@tvrusso that was scary. i let the AI make that comment. I had to enable a few things to let it respond and made sure to disable all permissions and approve that manually. still, I'm uninstalling that MCP server, lol.

tvrusso:
I'm getting some compiler warnings that should probably be addressed
```
../../Xastir/src/map_OSM.c: In function 'draw_OSM_tiles':
../../Xastir/src/map_OSM.c:1386:36: warning: format '%lu' expects argument of type 'long unsigned int', but argument 6 has type 'unsigned int' [-Wformat=]
 1386 |                           "%s/%d/%lu/%lu.%s", tileRootDir, osm_zl, row, col,
      |                                  ~~^                               ~~~
      |                                    |                               |
      |                                    long unsigned int               unsigned int
      |                                  %u
../../Xastir/src/map_OSM.c:1386:40: warning: format '%lu' expects argument of type 'long unsigned int', but argument 7 has type 'unsigned int' [-Wformat=]
 1386 |                           "%s/%d/%lu/%lu.%s", tileRootDir, osm_zl, row, col,
      |                                      ~~^                                ~~~
      |                                        |                                |
      |                                        long unsigned int                unsigned int
      |                                      %u
../../Xastir/src/map_OSM.c: In function 'draw_OSM_map':
../../Xastir/src/map_OSM.c:1806:3: warning: 'strncpy' output may be truncated copying 2053 bytes from a string of length 3999 [-Wstringop-truncation]
 1806 |   strncpy(image_info->filename, OSMtmp, MaxTextExtent);
      |   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

There are lots of other stringop-trunctation warnings throughout the code that have been ignored,  and this one is not new, so I'm not worried about it.

However  the printf issue is new and should be very easy to fix.

If you do fix it, please do so with an additional commit rather than a force push.  I have already pulled your branch and force-pushed amended commits mean I have to do more work to get the updates.

godfreja (approved):
Works on my Mac. I don't notice much of a performance difference, but I am running much more modern hardware than a Pi 2.
Code looked OK, but I am not much of an X11 programmer.

tvrusso:
I'm still going to give this a little more review, but will probably accept it in its current form.  Since I'm going to merge #377 right away (it is too simple and too obviously correct to wait on), please don't rebase and force-push this PR.  That is unnecessary and I can get this PR merged in due time without that.

tvrusso (approved):
Tested out extensively, including on a Raspberry Pi 3+ with 1920x1080p display.   The difference between the unmodified code and code after this PR is absolutely an obvious improvement in OSM tile rendering performance.

PR #378, opened 2026-05-06
Move online .geo map files into dedicated subdirectory

### LINKED ISSUE #355: Move WMS geo files out of top level directory
Back when geogratis and national map .geo files (all of which access WMS servers) were placed in the Xastir repo, they were put in the top level directory.  Because we were still using CVS at the time, moving files was problematic --- you could do it only by removing them from CVS and recommitting them in a different location, which lost all the history.  You could play repository games to move the actual repo files, but that broke the build if one tried to check out old versions.  So we left them there.

Later, after we moved to git, we just kept putting new online map .geo files into the top directory.

I was digging through old mailing list threads the other day and found one comment about this, about how odd it was to have these files a the top level.

Since it's super easy to move files in git and retain their history, we should create a new subdirectory for online geo files and move all of those files into it.

hjf:
This implements issue #355 by moving online map `.geo` files out of the repository top level into a dedicated `online-maps/` subdirectory.

## What changed
- Moved online map `.geo` files from repository root to `online-maps/`:
  - `USTigermap*.geo`
  - `WMS_USGS_*.geo`
  - `NWSRadar_*.geo`
  - `National.geo`, `Regional.geo`, `Sub_national.geo`, `Sub_regional.geo`
- Added `online-maps/Makefile.am` to keep existing install destinations under `maps/Online/*`.
- Updated top-level `Makefile.am`:
  - Added `online-maps` to `SUBDIRS`.
  - Kept only generated OSM tiled `.geo` files in top-level `online_DATA`.
- Updated `configure.ac` to generate `online-maps/Makefile`.

This preserves installed map layout while cleaning up source tree organization.

Fixes #355.


tvrusso (approved):
Exactly what had to be done.

PR #379, opened 2026-05-06
Fix signpost decode when altitude precedes signpost text (issue #259)

### LINKED ISSUE #259: Xastir incorrectly decodes signpost objects if altitude is present
Signpost objects may have speeds, courses, altitudes, and a three-character signpost text field.   If given, altitude in the form '/A=xxxxxx' is added to the comment field.  If signpost text is given, '{xxx}' is added to the comment field, always *AFTER* the altitude if it is given.

This text is initially created by `Setup_object_data` or `Setup_item_data`, and then fed into Xastir's packet processing pipeline, where it gets decoded by the stuff in db.c and inserted into the database.  Xastir recognizes the packet as coming from itself and flags it as an "own" object for transmission.

However, there is a bug in the logic of decoding these packets.  The scan of the comment field for signpost string (`extract_signpost`, called from `process_data_extension`) happens before the call to `extract_altitude` (called from `process_info_field`).   None of the functions called by `process_data_extension` work properly if the altitude string comes first.  These functions work by searching for the specific extension type at the beginning of the remaining part of the packet, and returning zero if they're not found.  This cannot work if altitude precedes them (which it shouldn't, if they're actually data extensions, because altitude belongs in the comment.

Image

But according to the spec, the signpost data is NOT part of the data extension.  It's supposed to be in the comment field.

Image

And according to the spec, the altitude string is allowed to appear *anywhere* in the comment field, and though it doesn't say so, it must also be the case that signpost data must be able to appear anywhere in the comment.

This issue causes incorrect parsing of incoming signpost objects formatted as Xastir creates them, and causes the signpost string to wind up stored in the comment, not in the signpost field of the DataRow.

I think the correct fix for this is to move `extract_signpost` out of `process_data_extension` (since this data is NOT a data extension) and into `process_info_field`, rewriting it so it does not care if some other type of data that also gets stuffed into the comment field precedes it.

We have to be careful, because signpost text strings look exactly like area object corridor strings ('{xxx}') but the corridor strings should only be there if the object (or item's) symbol is `\l` whereas the signpost strings should only be there for symbols `\m`.

This issue was exposed when I was trying to refactor Create_object_item_tx_string by extracting its pieces into simple functions.  Something about this bad decoding broke my hacking run.

c.f. #249 and #250

To reproduce:

1. Set yourself up to observe all outgoing packet transmissions
2. Create a signpost object or item using the right-click menu in the map window.   Select the "signpost" type.   Add a non-zero altitude in the altitude box and a one to three-character entry in the "Signpost Text" box.  Observe the  first outgoing packet, which will be formatted with '/A=xxxxxx{xxx}' with xxxxxx being the altitude and xxx being your signpost text.
3. right-click on the object and select Modify Object
4. You will observe that instead of having xxx in the Signpost Text field, '{xxx}' will be in the comment field.
5. You will also observe that the signpost text does not appear near the symbol for the signpost on the map.
6. If you remove the altitude and move the signpost text back to its rightful place in the dialog and modify the object, the signpost symbol will have the signpost text displayed near it, and further attempts to modify the object will have the signpost text in its correct place.

hjf:
## Problem

When a signpost object includes an altitude (`/A=xxxxxx`), Xastir fails to decode the signpost text (`{xxx}`). The `{xxx}` ends up stored in the comment field instead of `p_station->signpost`, so the signpost text is not displayed on the map and the Modify Object dialog shows it in the wrong place/format.

Root causes:

1. `extract_signpost()` only matched `{xxx}` at position 0 of the buffer, so `/A=xxxxxx{xxx}` was never matched.
2. `extract_signpost()` was called from `process_data_extension()`, **before** `extract_altitude()` in `process_info_field()` had a chance to strip the altitude string. Per the APRS spec, signpost text is a **comment field** item, not an APRS data extension.

## Fix (src/db.c)

**1. Rewrite `extract_signpost()`** to scan the entire string for `{xxx}` at any offset, similar to `extract_altitude()`. This makes the parser robust against other comment-field data preceding the signpost text.

**2. Remove the `extract_signpost()` call from `process_data_extension()`.**  Signpost text is not a data extension.

**3. Add the `extract_signpost()` call to `process_info_field()`**, guarded by the signpost symbol check (`aprs_type == '\\'` && `aprs_symbol == 'm'`). It runs *after* `extract_altitude()` has already stripped `/A=xxxxxx`, so packets with altitude + signpost text now decode correctly.

Area-object corridor strings (`{xxx}` for symbol `\l`) are handled by `extract_area()` in `process_data_extension()` and are unaffected.

## Reproduction (from issue)

1. Create a signpost object with a non-zero altitude and 1–3-char signpost text.
2. Observe the first transmitted packet — e.g. `...symbol\m/A=000100{ABC}`.
3. Right-click → Modify Object.
4. **Before**: signpost text field is empty; `{ABC}` appears verbatim in the comment field.
5. **After**: signpost text field shows `ABC`; map renders the signpost label correctly.

Fixes #259.

### DIFF
--- a/src/db.c
+++ b/src/db.c
@@ -7634,53 +7634,56 @@ int extract_omnidf(char *info, char *phgd)
 
 
 /*
- *  Extract signpost data from APRS info field: "{123}", an APRS data extension
+ *  Extract signpost data from APRS info field: "{123}"
  *  Format can be {1}, {12}, or {123}.  Letters or digits are ok.
+ *  Scans the entire string so altitude (/A=xxxxxx) or other comment
+ *  data may precede the signpost text without causing a parse failure.
  */
 int extract_signpost(char *info, char *signpost)
 {
-  int i,found,len,done;
+  int i, ofs, found, len, end;
 
-  //0123456
-  //{1}
-  //{12}
-  //{121}
+  found = 0;
+  len   = (int)strlen(info);
 
-  found=0;
-  len = (int)strlen(info);
-  if ( (len > 2)
-       && (info[0] == '{')
-       && ( (info[2] == '}' ) || (info[3] == '}' ) || (info[4] == '}' ) ) )
+  /* Scan for '{' followed by 1-3 chars and a closing '}' */
+  for (ofs = 0; !found && ofs < len - 2; ofs++)
   {
-
-    i = 1;
-    done = 0;
-    while (!done)                   // Snag up to three digits
+    if (info[ofs] == '{')
     {
-      if (info[i] == '}')         // We're done
+      if (   (ofs + 2 < len && info[ofs + 2] == '}')
+          || (ofs + 3 < len && info[ofs + 3] == '}')
+          || (ofs + 4 < len && info[ofs + 4] == '}') )
       {
-        found = i;              // found = position of '}' character
-        done++;
-      }
-      else
-      {
-        signpost[i-1] = info[i];
-      }
-
-      i++;
-
-      if ( (i > 4) && !done)      // Something is wrong, we should be done by now
-      {
-        done++;
-        signpost[0] = '\0';
-        return(0);
+        found = 1;
       }
     }
-    substr(signpost,info+1,found-1);
-    found++;
-    for (i=0; i<=len-found; i++)    // delete omnidf from data extension field
+  }
+
+  if (found)
+  {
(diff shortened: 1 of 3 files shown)


tvrusso (approved):
Nothing but net.   Fixes the issue *and* contains appropriate unit tests.

PR #381, opened 2026-05-16
Add global UTF-8 traffic toggle for APRS messages

### LINKED ISSUE #380: Add configurable UTF-8 handling for APRS message traffic
## Problem
Modern APRS clients (for example APRSdroid) commonly send message text as UTF-8.

In Xastir's Motif/Latin-1 UI paths this can appear as mojibake (for example ñ arriving as Ã±) in message windows such as the Send Message history view.

## Why this is needed
- Interoperability with modern clients that already send UTF-8
- Better readability for Western European languages
- Avoid regressions for legacy stations by making behavior configurable

## Proposed approach
Add a global config toggle that enables UTF-8 traffic handling for APRS message text:
- Incoming message parse/display paths: UTF-8 -> Latin-1 normalization for UI rendering
- Outgoing message path: Latin-1 UI input -> UTF-8 on-air payload
- Persist setting in config file for predictable behavior

## Scope and limitations
- Intended for Western European text compatibility (Latin-1 range)
- Not full Unicode rendering in Motif widgets
- No emoji support
- No CJK support
- Characters outside Latin-1 are replaced during UI normalization

A companion PR is prepared with this implementation and build/test updates.

hjf:
## Summary
This PR adds configurable UTF-8 handling for APRS message traffic to improve interoperability with modern clients such as APRSdroid.

It introduces a global config option that controls both:
- incoming message normalization for display in legacy Motif/Latin-1 UI paths, and
- outgoing message encoding so transmitted payloads can be UTF-8.

Closes: #380

## What changed
- Added global runtime/config variable `traffic_utf8_enabled`.
- Persisted setting in `xastir.cnf` as `TRAFFIC_UTF8` (load/save).
- Gated incoming conversion in message ingest/display code paths.
- Gated outgoing conversion in message output path.
- Updated unit-test stubs to include the new global symbol used by `db.c` test links.

## Behavior
- `TRAFFIC_UTF8=1` (default):
  - Incoming: normalize UTF-8 message text for Latin-1 UI display.
  - Outgoing: encode Latin-1 UI input to UTF-8 before transmit.
- `TRAFFIC_UTF8=0`:
  - Preserve legacy byte behavior (no UTF-8 parse/send conversions).

## Limitations
- This is not full Unicode UI support.
- Scope is Western European language compatibility (Latin-1 range).
- No emoji support.
- No CJK support.
- Characters outside Latin-1 are replaced during UI normalization.

## Validation
- Full build/install script completed successfully (`./build.sh`, exit 0).
- Addressed unit-test link dependencies by adding stub definitions for the new global.

### DIFF
--- a/config/language-Dutch.sys
+++ b/config/language-Dutch.sys
@@ -430,6 +430,7 @@ WPUPCFD030|Zet positie duplicaat controle uit||
 WPUPCFD031|Laad voor gedefinierde objecten uit bestand||
 WPUPCFD032|My trails in one color||
 WPUPCFD033|ALTNET:||
+WPUPCFD034|Verkeer als UTF-8 interpreteren|U|
 #
 # PopUp "Configure - Timing"
 WPUPCFTM01|Configureer tijdsinstelling|||
--- a/config/language-English.sys
+++ b/config/language-English.sys
@@ -422,6 +422,7 @@ WPUPCFD030|Disable Posit Dupe-Checks||
 WPUPCFD031|Load predefined objects from file||
 WPUPCFD032|My trails in one color||
 WPUPCFD033|ALTNET:||
+WPUPCFD034|Parse traffic as UTF-8|U|
 #
 # PopUp "Configure - Timing"
 WPUPCFTM01|Configure Timing||
--- a/config/language-French.sys
+++ b/config/language-French.sys
@@ -423,6 +423,7 @@ WPUPCFD030|Désactiver identif. de duplic. posit.||
 WPUPCFD031|Charger des objets prédéfinis à partir d'un fichier||
 WPUPCFD032|Mon parcour en une seule couleur||
 WPUPCFD033|ALTNET:||
+WPUPCFD034|Analyser le trafic en UTF-8|U|
 #
 # PopUp "Configure - Timing"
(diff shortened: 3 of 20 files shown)


tvrusso:
Your PR looks like it creates a new config variable for xastir.cnf, but does it add any way of setting it other than by hand-editing the config file?

hjf:
> Your PR looks like it creates a new config variable for xastir.cnf, but does it add any way of setting it other than by hand-editing the config file?

added UI for this and translations. I'm not sure if we should enable this by default. It's unlikely to affect english users. And it should only affect non-english users if they send a weird multibyte string which is likely not to be a valid ISO-8859 character they actually meant to send, so those will still behave as 8859.

Should be transparent.

tvrusso (approved):
Tested out, and I can correctly enter accented characters in the new version and have it transmitted as UTF8, and if I receive it on an Xastir instance without this fix I get mojibake.

Further, prior to this PR, trying to enter accented characters into a message sent them in the Latin-1 encoding, which was wrong.   Now they're properly sent in UTF-8.  

So this PR gets both encode and decode right.

The language translation files are all updated.

I don't think there's any harm in setting it as default.   It does not appear to break anything if one *isn't* using UTF-8 characters.

PR #391, opened 2026-07-21
Fix uninitialized config paths on save failure

simonbleher:
### Problem
In `save_data()` (`src/xa_config.c`), the buffers `config_file` and
`config_file_bak1..4` are only populated (via `get_user_base_dir()`)
**inside** the `if (fout != NULL)` block. When `fopen()` of the temp
config file fails (disk full, permission denied, read-only home, …),
execution falls into the `else` branch, which then uses those buffers
uninitialized:

- `fprintf(stderr, "...%s\n", config_file);`
- `rename(config_file_bak1, config_file);`

This is undefined behaviour: the recovery code that is meant to restore
the previous configuration operates on uninitialized stack contents as
pathnames, so it can fail silently or rename unintended files.

### Fix
Move the `get_user_base_dir()` calls that build these paths ahead of the
`fopen()`, so both the success path and the error/recovery path see
valid pathnames. No behavioural change on the normal path.

### Testing
`./bootstrap.sh && ./configure && make` builds cleanly (0 errors);
`xa_config.o` compiles without new warnings. cppcheck no longer reports
the `uninitvar` warnings at the affected lines.

### DIFF
--- a/src/xa_config.c
+++ b/src/xa_config.c
@@ -545,6 +545,20 @@ void save_data(void)
   get_user_base_dir(CONFIG_FILE_TMP, config_file_tmp,
                     sizeof(config_file_tmp));
 
+  // Fill in the config and backup file paths up front.  These must be
+  // initialized before the fopen() below so that the error/recovery
+  // path (the "else" branch when fopen() fails) doesn't reference
+  // uninitialized buffers when logging and renaming files.
+  get_user_base_dir(CONFIG_FILE, config_file, sizeof(config_file));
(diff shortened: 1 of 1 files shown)


tvrusso:
Another one that has been wrong since 2004.   The problematic code was introduced in commit e26da87, which placed the creation of the backup file names into the if( fout != NULL) block.  Prior to that there was only one backup file name and it had been getting set before the fopen.

This is obviously a correct fix and I'm merging it.

PR #395, opened 2026-07-22
Bound sscanf field widths in GEO file parsing

simonbleher:
While reading through `map_geo.c` I noticed that `draw_geo_image_map()`
  extracts tokens from GEO file lines with a bare `%s` and no field width.

  Each line is read with `get_line(f, line, MAX_FILENAME)`, so `line[]` can
  hold up to 1999 characters. Six `sscanf()` calls then copy a token out of
  that line into buffers that are smaller than the longest possible token:

  | buffer | size | read from |
  | --- | --- | --- |
  | `OSMtileExt[MAX_OSMEXT]` | 10 bytes | `line + 9` (TILE_EXT) |
  | `OSMstyle[MAX_OSMSTYLE]` | 1000 bytes | `line + 13`, `line + 14` |

  So a GEO file with a long `TILE_EXT` or `OSMSTATICMAP` / `OSM_TILED_MAP`
  token overflows these stack buffers.

  `fileimg[]` and `tileCache[]` are `MAX_FILENAME+1` and happen to be large
  enough today, but only because of the size of `line[]` rather than their
  own, so I bounded those too.

  The widths match the destination buffers and follow the style already used
  a few lines above for `DATUM` (`"%8s"`) and `PROJECTION` (`"%256s"`).

  ### Verification

  I reproduced the `TILE_EXT` case in a standalone program using the same
  buffer sizes and the same read path. With the current code AddressSanitizer
  reports:

  ```
  ERROR: AddressSanitizer: stack-buffer-overflow
  WRITE of size 1501 at 0x... thread T0
      #0 scanf_common
    This frame has 3 object(s):
      [32, 42) 'OSMtileExt'
  ```

  With the widths added it runs clean and the token is truncated at 9 and 999
  characters as expected.

  I also built the full tree (bootstrap/configure/make) with no new warnings.
  The two remaining warnings in this file are pre-existing and are at line
  1138, not at any of the lines I touched.

  `make check` passes locally as well (all 252 tests successful), though the
  suite does not currently cover `map_geo.c`, so that is a regression check
  rather than direct coverage of this change.

### DIFF
--- a/src/map_geo.c
+++ b/src/map_geo.c
@@ -718,7 +718,7 @@ void draw_geo_image_map (Widget w,
       (void)get_line (f, line, MAX_FILENAME);
       if (strncasecmp (line, "FILENAME", 8) == 0)
       {
-        if (1 != sscanf (line + 9, "%s", fileimg))
+        if (1 != sscanf (line + 9, "%2000s", fileimg))
         {
           fprintf(stderr,"draw_geo_image_map:sscanf parsing error\n");
         }
@@ -740,7 +740,7 @@ void draw_geo_image_map (Widget w,
         }
       }
       if (strncasecmp (line, "URL", 3) == 0)
-        if (1 != sscanf (line + 4, "%s", fileimg))
+        if (1 != sscanf (line + 4, "%2000s", fileimg))
         {
           fprintf(stderr,"draw_geo_image_map:sscanf parsing error\n");
         }
@@ -814,7 +814,7 @@ void draw_geo_image_map (Widget w,
         OSMserver_flag = 1;
         if (strlen(line) > 13)
         {
-          if (1 != sscanf (line + 13, "%s", OSMstyle))
+          if (1 != sscanf (line + 13, "%999s", OSMstyle))
           {
             fprintf(stderr,"draw_geo_image_map:sscanf parsing error for OSM style.\n");
           }
@@ -826,7 +826,7 @@ void draw_geo_image_map (Widget w,
         OSMserver_flag = 2;
         if (strlen(line) > 14)
         {
-          if (1 != sscanf (line + 14, "%s", OSMstyle))
+          if (1 != sscanf (line + 14, "%999s", OSMstyle))
           {
             fprintf(stderr,"draw_geo_image_map:sscanf parsing error for OSM style.\n");
           }
@@ -877,7 +877,7 @@ void draw_geo_image_map (Widget w,
         {
           if (strlen(line) > 9)
           {
-            if (1 != sscanf (line + 9, "%s", tileCache))
+            if (1 != sscanf (line + 9, "%2000s", tileCache))
             {
               fprintf(stderr,"draw_geo_image_map:sscanf parsing error for TILE_DIR\n");
             }
@@ -888,7 +888,7 @@ void draw_geo_image_map (Widget w,
         {
           if (strlen(line) > 9)
           {
-            if (1 != sscanf (line + 9, "%s", OSMtileExt))
+            if (1 != sscanf (line + 9, "%9s", OSMtileExt))
             {
               fprintf(stderr,"draw_geo_image_map:sscanf parsing error for TILE_EXT\n");
             }


tvrusso:
Well spotted.   The handling of strings in Xastir isn't very consistent and in many cases unsafe things are being done.

Before I review and merge this (which might take a couple of days due to other conflicts), could you let me know what OS you're using?  Have you actually tried to use map file names that overrun these buffers in Xastir, or have you just spotted a *potential* problem that could bite us in the future?  

I am somewhat concerned about hard coding the lengths in the sscanf format strings.   Having hard coded lengths in the format string negates the value of having the maximum lengths of buffers set in "#define" macros, and can be a maintenance headache.   The more I think of that, the less I like it, and the more I think a better solution is possible.

In each of these cases, all that is happening is that everything on the line after the keyword is being copied into a buffer, which makes sscanf seem like a wasteful choice in the first place.  

Perhaps there is a cleaner solution that can copy substrings out of "line" and into the destination buffers that isn't relying on sscanf.   I would think that strncpy with appropriate arguments would be a better choice than sscanf --- especially since no actual data type conversion is being done, just straight copies of characters from the starting point to the end of line.   I will have to think about that before accepting this PR, and perhaps Jason can give his thoughts as well.  

If you want to spend more time on it, maybe you could investigate that option, since it will be a few days before I can get to it.

For Jason's reference:  I should note that unlike the issue fixed in #385, which was triggered by the fact that strict POSIX compliance means that "%Nc" in an sscanf really means "read *exactly* N characters" (causing a failure on a very recent OpenSUSE system), POSIX dictates that "%Ns" means "read up to N characters" and so this PR should NOT be a problem and it is almost certainly safe to merge, modulo the concern about having hard coded lengths in the format string.

It is unfortunate that our unit testing system isn't able to deal with the map code, but because those are so tightly tied to the graphics system it's not easy to add them without breaking up the files so that the non-graphics code can be unit tested.   This would be a perfect thing to have tested there.

tvrusso:
On reflection, using "xastir_snprintf" (which is  in almost all cases just a macro that resolves to "snprintf") to copy the characters safely and with proper null termination, without the possibility of buffer overrun, is probably better than using strncpy (which can fail to terminate strings in some cases) *or* sscanf (which would require hardcoding buffer sizes into format strings).

Handling the cases of failure would take a bit of reworking, though.   snprintf returns the number of characters that *would* be printed to the buffer if size were not limited, so perhaps checking that the return value is greater than the buffer size would be a good flag that there is a failure.   Have to think about that before jumping in and doing it.

simonbleher:
Thanks for the detailed look.

I'm on Ubuntu 22.04.5 LTS, gcc 11.4.0, kernel 6.8.

To be clear about how I found it: this is a potential problem, not one I
hit in normal use. I was reading through the file, noticed the missing
widths, and then confirmed it was reachable by building a standalone
program with the same buffer sizes and the same read path and feeding it
an over-long TILE_EXT line. That is where the ASan output in the
description comes from. I have not seen a real map file trigger it.

I think your concern about the hard coded lengths is fair, and I would
rather fix it properly than leave that in.

One thing worth noting about replacing sscanf with strncpy: "%s" does
slightly more than copy the rest of the line. It skips leading
whitespace and stops at the first whitespace character. get_line() only
strips CR/LF, not trailing spaces, so a hand-edited line like
"TILE_EXT png   " would give "png" today but "png   " with a plain
strncpy to end of line. Probably not what we want.

So what I would suggest is a small helper in util.c along the lines of

    char *copy_token(char *dest, size_t dest_size, const char *src);

which skips leading whitespace, copies up to the next whitespace or end
of string, and always bounds the copy by dest_size. Call sites then pass
sizeof(dest), so the #define values stay authoritative and nothing is
hard coded in map_geo.c.

That also has the side benefit of being unit testable. util.c is already
covered by tests/util_tests.at, so the bounding behaviour could be tested
directly even though the map code itself can't be.

Happy to put that together as a revision of this PR while you're busy, if
you and Jason think it's a reasonable direction. Or if you'd prefer to
keep this PR as the minimal bounding fix and do the cleanup separately,
that works too.

tvrusso:
I think that fixing it properly using a unit-testable utility function would be an *excellent* way to handle this, and if you're willing to revise your PR (or close this one and open a new one), I would love to see it.   Making a clean, reusable function is the best way forward, rather than having scattered, clumsy, and unsafe bits of code duplication throughout the file.

I would suggest the following:
 - Open an issue report documenting the problem you found
 - open a new PR fixing the issue carefully, placing the helper function in util.c
    - include new util.c unit test case mapping out correct behavior both for valid input and malformed input

If you'd rather just document the problem in an issue report and leave the fix for us, that's fine, but if you want to take the time to fix it in a clean way, please feel free.   We welcome the help and I thank you for your interest in improving the Xastir code base.

simonbleher:
Closed in favour of #397 , which takes the util.c helper approach we
discussed here instead of hard coding the widths in the format strings.

PR #397, opened 2026-07-22
Add copy_token() and use it in GEO file parsing

### LINKED ISSUE #396: Unbounded token reads in GEO file parsing can overflow stack buffers
`draw_geo_image_map()` in `src/map_geo.c` reads each line of a GEO file
into `line[]` with `get_line(f, line, MAX_FILENAME)`, so a line can hold
up to 1999 characters. Several `sscanf()` calls then extract a token with
a bare `"%s"` and no field width, into buffers that are smaller than the
longest token a line can carry:

| buffer | size | read from |
| --- | --- | --- |
| `OSMtileExt[MAX_OSMEXT]` | 10 bytes | `line + 9` (TILE_EXT) |
| `OSMstyle[MAX_OSMSTYLE]` | 1000 bytes | `line + 13`, `line + 14` |

A GEO file with a long `TILE_EXT` or `OSMSTATICMAP` / `OSM_TILED_MAP`
token overflows these stack buffers.

`fileimg[]` and `tileCache[]` are `MAX_FILENAME+1` and happen to be large
enough today, but only because of the size of `line[]`, not because of
their own size.

### How I found it

Reading the source. I have not seen a real map file trigger this. I
confirmed it is reachable by building a standalone program with the same
buffer sizes and the same read path and feeding it an over-long
`TILE_EXT` line. AddressSanitizer reports:

```
ERROR: AddressSanitizer: stack-buffer-overflow
WRITE of size 1501 at 0x... thread T0
    #0 scanf_common
  This frame has 3 object(s):
    [32, 42) 'OSMtileExt'
```

Environment: Ubuntu 22.04.5 LTS, gcc 11.4.0, kernel 6.8.

### Note on the fix

Simply adding field widths (`"%9s"` etc.) removes the overflow but writes
each buffer size into a string literal, so the `#define` values are no
longer authoritative. Discussed in #395; a cleaner approach using a
unit-testable helper in `util.c` is in the PR that follows.

simonbleher:
Fixes #396. This replaces #395 with the cleaner approach.

### The helper

New function in `util.c`:

```c
int copy_token(char *dest, size_t dest_size, const char *src);
```

It skips leading whitespace, copies up to the next whitespace character
or end of string, always NUL terminates, and never writes more than
`dest_size` bytes. The copy itself is done with `xastir_snprintf()` as
you suggested, so termination is guaranteed and there is no `strncpy()`
edge case to worry about.

The return value is the length the token would have needed, so callers
can tell the two failure modes apart:

* `0` means no token was found
* `>= dest_size` means the token was truncated

Call sites pass `sizeof(dest)`, so the `#define` values stay
authoritative and no buffer size is written into a format string.

### Call sites

All eight string token extractions in `draw_geo_image_map()` now use it.
That includes the `DATUM` and `PROJECTION` reads, which were already
bounded with `"%8s"` and `"%256s"` but had the same hard coded length
problem. After this change there is no buffer size in any format string
in the file.

The `sscanf()` calls doing numeric conversions (`%d`, `%f`, `%lu`) are
unchanged, since that is what `sscanf()` is actually for.

I also dropped the word "sscanf" from the diagnostics at the converted
sites, since it no longer describes what happens there. The four
diagnostics that share the same wording but still belong to real
`sscanf()` calls were left alone.

### Tests

Eight new cases in `tests/util_tests.at`, covering valid and malformed
input as you asked:

| case | behaviour |
| --- | --- |
| plain token | copied unchanged |
| leading whitespace | skipped |
| trailing whitespace | copy stops at it |
| exact buffer fit | not reported as truncation |
| oversized token | truncated, full length returned |
| empty input | returns 0 |
| whitespace-only input | returns 0 |
| NULL / zero size | rejected without crashing |

`make check` goes from 252 to 260 tests, all passing. I also rebuilt
`tests/test_util` with `-fsanitize=address` and ran the eight cases
against the real function; the truncation case pushes a 10 character
token into a 4 byte buffer and ASan reports nothing.

Full tree builds clean with no new warnings.

### DIFF
--- a/src/map_geo.c
+++ b/src/map_geo.c
@@ -718,9 +718,9 @@ void draw_geo_image_map (Widget w,
       (void)get_line (f, line, MAX_FILENAME);
       if (strncasecmp (line, "FILENAME", 8) == 0)
       {
-        if (1 != sscanf (line + 9, "%s", fileimg))
+        if (0 == copy_token(fileimg, sizeof(fileimg), line + 9))
         {
-          fprintf(stderr,"draw_geo_image_map:sscanf parsing error\n");
+          fprintf(stderr,"draw_geo_image_map:parsing error\n");
         }
         if (fileimg[0] != '/' )   // not absolute path
         {
@@ -740,9 +740,9 @@ void draw_geo_image_map (Widget w,
         }
       }
       if (strncasecmp (line, "URL", 3) == 0)
-        if (1 != sscanf (line + 4, "%s", fileimg))
+        if (0 == copy_token(fileimg, sizeof(fileimg), line + 4))
         {
-          fprintf(stderr,"draw_geo_image_map:sscanf parsing error\n");
+          fprintf(stderr,"draw_geo_image_map:parsing error\n");
         }
 
       if (n_tp < 2)       // Only take the first two tiepoints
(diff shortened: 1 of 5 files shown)


tvrusso:
Thank you for getting on this so quickly.   As noted, I won't have time to review it today.   But it'll be a top priority as soon as my plate clears.

tvrusso (approved):
I think this is a solid contribution that fixes a long-standing bit of relatively careless string handling in a clean, modular, and tested manner.

Thank you for another excellent contributions and regards to the elder Bleher.

PR #399, opened 2026-07-25
Expire and prune stale entries from wx alert log cache

### LINKED ISSUE #125: Weather alerts cached too long
Weather alerts are kept with no time/date-stamp in several files. If Xastir is shut down for a period of time and then started up again, it may bring weather alerts back in and display them after they should have been expired. We should write timestamps into the files, only load them back in if they're within a specific window, and actively delete expired ones from the cache files.
~/.xastir/logs/wx_alert.log
~/.xastir/logs/wx_alert.log.1
~/.xastir/logs/wx_alert.log.2

simonbleher:
Fixes #125.

load_wx_alerts_from_log_working_sub() already skipped log files and individual entries older than the 15 day expiration window when loading, but never removed anything from disk: a fully expired log file was left in place forever, and expired entries mixed into an otherwise-current file were skipped on load but never dropped, so they accumulated indefinitely between size-triggered rotations.

Now a fully expired file is removed outright, and files within the window are rewritten to keep only entries that are still current.

Adds a unit test suite (test_log_utils) covering whole-file removal, pruning of stale entries while keeping fresh ones, keeping all entries when nothing has expired, and handling a missing file.

### DIFF
--- a/src/log_utils.c
+++ b/src/log_utils.c
@@ -176,7 +176,9 @@ void load_wx_alerts_from_log_working_sub(time_t time_now, char *filename)
   int file_age;
   int expire_limit;   // In seconds
   char line[MAX_LINE_SIZE+1];
+  char tmp_filename[MAX_FILENAME];
   FILE *f;
+  FILE *ftmp;
 
 
   expire_limit = 60 * 60 * 24 * 15;   // 15 days
@@ -194,7 +196,8 @@ void load_wx_alerts_from_log_working_sub(time_t time_now, char *filename)
 
   if ( file_age > expire_limit)
   {
-//        fprintf(stderr,"Old file: %s, skipping...\n", filename);
+    fprintf(stderr,"Old wx alert log file, removing: %s\n", filename);
+    (void)remove(filename);
     return;
   }
 
@@ -225,6 +228,13 @@ void load_wx_alerts_from_log_working_sub(time_t time_now, char *filename)
     return;
   }
 
+  xastir_snprintf(tmp_filename, sizeof(tmp_filename), "%s.tmp", filename);
+  ftmp = fopen(tmp_filename, "w");
+  if (!ftmp)
+  {
+    fprintf(stderr,"Couldn't create %s, wx alert log won't be pruned\n", tmp_filename);
+  }
+
   while (!feof(f))    // Read until end of file
   {
 
@@ -241,6 +251,7 @@ restart_sync:
     {
       time_t line_stamp;
       int line_age;
+      char timestamp_line[MAX_LINE_SIZE+1];
 
       if (strlen(line) < 3)   // Line is too short, skip
       {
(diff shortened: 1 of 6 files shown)


tvrusso:
I pulled this branch and tried running it with wx_alert logging turned on.   I found that it did indeed prune the weather alert log, but then created a new log with incorrect contents.

Prior to the fix, the alert log entries looked like this:
```
# 1785013017  Sat Jul 25 14:56:57 MDT 2026
CRPFLS>APRS::NWS-WARN :261030z,FLOOD,TXC283{P00AA
# 1785013017  Sat Jul 25 14:56:57 MDT 2026
CRPFLS>APRS::NWS_WARN :261030z,FLOOD,TXC283{P00AA
# 1785013017  Sat Jul 25 14:56:57 MDT 2026
LBFSVR>APRS::NWS-WARN :252145z,SVR_STORM,NEC031{w40AA
# 1785013018  Sat Jul 25 14:56:58 MDT 2026
LBFSVR>APRS::NWS_WARN :252145z,SVR_STORM,NEC031{w40AA
```

I had an old log with very old entries in it and it was rewritten by this new code to:
```
# 1785012777  Sat Jul 25 14:52:57 MDT 2026

CRPFLS
# 1785012777  Sat Jul 25 14:52:57 MDT 2026

CRPFLS
# 1785012777  Sat Jul 25 14:52:57 MDT 2026

ABQFFW
# 1785012777  Sat Jul 25 14:52:57 MDT 2026

ABQFFW
# 1785012778  Sat Jul 25 14:52:58 MDT 2026

ABRSVR
# 1785012778  Sat Jul 25 14:52:58 MDT 2026

ABRSVR
# 1785012778  Sat Jul 25 14:52:58 MDT 2026

ABQFFW
# 1785012778  Sat Jul 25 14:52:58 MDT 2026

ABQFFW
```

That is, all the information about the alerts was discarded and extra spaces were introduced.

Unfortunately, I had not saved the old log file and so I won't be able to reproduce this failure until my log file has stale data again.

simonbleher:
Thanks for testing this and catching it — reproduced and fixed, two separate bugs stacked on top of each other:

1. `decode_ax25_line()` tokenizes its `line` argument in place with `strtok(line, ">")`, truncating the buffer at the first `>`. The pruning code called `decode_ax25_line(line, ...)` and then wrote that same, now-truncated buffer to the pruned file — hence entries getting cut down to just the sender callsign. Fixed by snapshotting the packet into a separate buffer before decoding, and writing the snapshot instead.

2. `fetch_file_line()` excluded a trailing CR from the returned line but not a trailing LF — the LF was getting appended to the buffer and only *then* null-terminated, so every line came back with an embedded trailing `\n`. That, combined with the `\n` already in the pruning code's format string, produced the blank line before each entry. Fixed to drop the LF the same way CR was already being dropped.

Added a regression test (`prune_preserves_full_packet_content`) that reproduces this exact corruption with realistic `CALL>PATH:data`-style content, and updated the other tests to use similar content instead of delimiter-free placeholder strings, since those didn't actually exercise `decode_ax25_line`'s mutation. Also made the `decode_ax25_line()` test stub mimic the real function's `strtok()` behavior instead of being a no-op, so this class of bug gets caught in CI going forward rather than only by hand-testing.

Pushed as a new commit on this branch (09eb3483). All 265 tests pass, CI is green on all four platforms.

If you still have that old log file with stale data lying around, I'd still be curious whether the file-level and line-level expiration windows themselves are working as expected for you — but the corruption you saw should be fully gone now.

tvrusso:
Much better.   This time, before testing it out, I saved my old log file.   I also temporarily changed the expire limit to 12 hours instead of 15 days, to force a prune.    The result was that the pruned file differed from the saved file *only* by the removal of expired alerts, with all unexpired alerts copied into the pruned file verbatim.

Thank you again for your efforts.  I'll be merging this one.

